Issue #39, opened 2021-02-26
Start vitess servers with pprof option


frouioui:
### Context

This issue will start a list of upcoming features and enhancements that enable profiling of Vitess throughout benchmarks.

Luckily, @vmg has implemented vitessio/vitess#7496 easing the profiling of Vitess.

### Description

This initial issue requests the addition of the `-pprof` flag to Vitess servers. These changes will have to take place within our Ansible configuration and also the CLI.

Let's start with the CLI, we want the activation of `-pprof` flag to be dynamic, thus a new CLI flag will be required. Moreover, additional configuration regarding profiling (path, etc) could be used.

Finally, Ansible, the flags we pass down to (for instance) `vtgate` will need to be changed. As we can see in the snippet  below, flags are listed one by one using Jinja and ultimately we will need to make `-pprof` dynamic as well:
https://github.com/vitessio/arewefastyet/blob/05bb6baf421e0cf3fd21412f2079282a1737d4cc/ansible/roles/vtgate/templates/vtgate%40.service.j2#L24-L41

PR #46, opened 2021-03-03
Start vitess servers with pprof option

frouioui:
This pull request contains the changes requested by #39. Vitess servers are now started with the pprof flag, and the profile files are fetched from the server to the local environment once a task is completed.

- An Ansible dependence had to be changed from `node-exporter` to `node_exporter`.
- Unit tests were added and corrected previous tests to reflect new changes.
- In the process of adding `--tasks-pprof` flag, the way to deal with a large number of arguments got simplified.
- Addition of the `-pprof` vitess serverenv flag inside Ansible configuration.


Akilan1999 (approved):
Amazing !

PR #97, opened 2021-03-26
Rolling update server

### LINKED ISSUE #42: Rolling update of the server
Setup and automate the deployment of the benchmark server on new releases and commits to master. This should be achieved through GitHub Actions.

Arewefastyet HEAD = deployed server HEAD

frouioui:
## Description

This pull request adds the changes required to automatically roll out new versions of the UI. The `supervisord` configuration in `./supervisor` contains fewer files and services.

The `server` and `reverse-proxy` remain, though they both got modifications to fit with the new code. Additionally, a script is added to perform the update of supervisord.

Finally, a new workflow that will deploy changes to the server on each push against master is added. This workflow will be executed only on master. I tested out the configuration on my own fork ([here](https://github.com/frouioui/arewefastyet/runs/2202339397?check_suite_focus=true)).

## Actions performed on the server

The new workflow performs the following set of commands:

```sh
cd $HOME/arewefastyet

# get rid of changes not staged, for instance, go.sum changes after go build resetting removes conflicts
git reset --hard FETCH_HEAD

# clean untracked files such as binaries
git clean -fd

# fetch the newer version
git fetch origin master

# checkout to newer version
git checkout FETCH_HEAD

# build the binary
/usr/local/go/bin/go build ./go/main.go

# update supervisor configuration and restart the service
./supervisor/update.sh
```

Resolves #42.


Akilan1999 (approved):
Awesome work on the workflows !

PR #108, opened 2021-03-30
Fix: command microbench run not returning status 0 when pkg path is incorrect

### LINKED ISSUE #88: Command microbench run not returning status 0 when pkg path is incorrect
### Description

When we feed the `microbench run` command with a wrong `pkg path` the program will still exit `0`. The command below can reproduce the bug:

```sh
arewefastyetcli microbench run invalid_path output.txt --config /tmp/config.yaml

echo $?
0
```

frouioui:
## Description

This pull request fixes the bug described in #88. The bug was that no output nor error code was produced by the CLI's `microbench run` command when feeding it a wrong path for the packages. In short, the following command behaved like everything went well:

```sh
arewefastyetcli microbench run invalid_path output.txt --config /tmp/config.yaml
```

## How to fix it?

In the `./go/tools/microbench` package's `MicroBenchmark` function, we load the packages we want to use. Loading packages with `packages.Load()` does not necessarily return a non-nil error if the given packages are incorrect.

```go
// below, the package.Load function's signature

func Load(cfg *Config, patterns ...string) ([]*Package, error)
```

However, the array of packages that we get (`[]*packages.Package`) will be populated with the packages we gave in the `patterns` argument of the function `packages.Load()`. The elements of the array contain a field `Errors` (an array of `package.Error`).

I then decided to verify this field for each package we treat. If the field contains one or more error, I append it to a slice of `error`, which eventually gets transform into a single `error` that we will return to the caller of function `MicroBenchmark`.

## Deal with errors

Another side of this issue was that errors were not handled in the `MicroBenchmark` function, which would instead `panic` or `log.Fatal`. This behavior was removed, instead, we return an error.

This allows us to treat error even higher in the stack, "microbench run" Cobra's command functions are now dealing with the error.

## Concatenation of errors

As mentioned earlier, in order to warn the user of all the errors we encountered with the packages and have a higher level of verbose, all the errors from `package.Package` are appended to a slice, which is then transformed into a single `error`. This is achieved by a newly added package in `./go/tools/errors`. This package will contain the functions and methods we need to ease the work we do around errors.

## How to reproduce and verify the fix

A new unit test was added for our `MicroBenchmark` function, the test contains a single test case and verifies that the function behaves correctly when given a wrong package path.

```sh
go test -v ./go/... -run ^TestMicroBenchmark$
```

Additionally, the fix can be verified by running:

```sh
go run ./go/main.go microbench run INVALID output.txt --config config/config.yaml 2>&1 >/dev/null | head -n 2
```

Which should output the following content:

```
Error: invalid package parsing:
package INVALID is not in GOROOT (/usr/local/go/src/INVALID)
```

## Related issues

Fixes #88.


go/tools/microbench/microbench.go
```diff
@@ -159,23 +166,37 @@ func MicroBenchmark(cfg MicroBenchConfig) {
...
 			}
 		}
 		fmt.Println()
 	}
+	return nil
 }
 
-func findBenchmarks(loaded []*packages.Package) []benchmark {
-	var benchmarks []benchmark
+func findBenchmarks(loaded []*packages.Package) (benchmarks []benchmark, errs []error) {
```
harshit-gangal, line 186: you can return a single error from here itself and does not need to concatenate outside.

go/tools/errors/errors.go
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * /
+ */
+
+package errors
+
+import "errors"
```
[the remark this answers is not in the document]
frouioui: Good call! Followed the recommendation and replaced it with `mutlierr`. Pushed in 767a8e60041cb2e6dff193456c17de9e016eaedb

go/tools/microbench/microbench.go
```diff
@@ -190,7 +207,10 @@ func findBenchmarks(loaded []*packages.Package) []benchmark {
 			}
 		}
 	}
-	return benchmarks
+	if len(multierr.Errors(err)) > 0 {
+		return nil, err
+	}
+	return benchmarks, nil
```
harshit-gangal, line 213: checking err != nil should be reasonable here.

harshit-gangal (commented):
I see there are fmt.println used in the code. This can be replaced with using log package.

go/tools/microbench/microbench.go
frouioui: Agreed, fixed in 467093b014ef9bfc7f17b0024d07417caf3e0740.

frouioui:
> I see there are fmt.println used in the code. This can be replaced with using log package.

You're right. Solved it in f406c65. This is a "quick-fix", in the future I would like to integrate some stdout/stderr io.Writer directly within `MicroBenchConfig` or else (the package itself, a `MicroBench` mother struct, ...). I am already in the process of integrating such thing in the `exec.Exec` struct, which you can see [here](https://github.com/frouioui/arewefastyet/commit/534316c9281424bd135258f22f8a580f5afa3696).

I replaced the `fmt.Print*` to `log.Print*` throughout three packages: `microbench`, `cmd`, and `infra`.

harshit-gangal (approved):
for now it is ok to use log.print.. later you would be more interested to log and classify them as INFO, WARN, ERROR.

PR #109, opened 2021-03-31
Macrobenchmarks (OLTP & TPCC) Graphs in Go WebUI

### LINKED ISSUE #91: Integrating add graphs for OLTP and TPCC runs
- [ ] Add OLTP and TPCC graphs 
- [ ] Change Query to displays commit for past 30 days

frouioui:
## Description

This pull request introduces a new package: `macrobench`. This package will be the root responsible for dealing with macro benchmarks such as OLTP and TPCC. A `results.go` file is in the package to fulfill a job like the one `microbench` package has, which is to deal with all the results: declare the proper structs we need, select the data from MySQL, etc.

By adding this package, we are able to reintroduce the graphs in the web UI. The graphs are now linked to the `macrobench` package.

A few trivial change were performed:

- The query connection string to MySQL contains a `parseTime=true` parameter, it was required to parse the `DateTime` columns into go's `*time.Time`.
- The graphs's Y-axis originates from 0.

## How to test

```sh
go run ./go/main.go --config config/config.yaml web

# and then open browser on localhost:<the port you defined in config>
```

## Related issues

Resolves #91.


go/server/handlers.go
```diff
@@ -33,8 +34,22 @@ func (s *Server) informationHandler(c *gin.Context) {
 }
 
 func (s *Server) homeHanlder(c *gin.Context) {
```
harshit-gangal, line 36: nit: `homeHanlder` -> `homeHandler`

go/server/handlers.go
```diff
@@ -33,8 +34,22 @@ func (s *Server) informationHandler(c *gin.Context) {
...
+	if err != nil {
+		log.Println(err.Error())
+		return
+	}
+
+	tpccData, err := macrobench.GetResultsForLastDays(macrobench.TPCC, "webhook", 31, s.dbClient)
+	if err != nil {
+		log.Println(err.Error())
+		return
+	}
```
harshit-gangal, line 47: can some error message be shown on the UI if the data is not retrieved?

go/tools/macrobench/macrobench.go
```diff
@@ -0,0 +1,20 @@
+/*
```
harshit-gangal, line 1: this seems to be an empty file and can be removed.

go/tools/macrobench/results.go
```diff
@@ -0,0 +1,125 @@
...
+			&res.Result.Errors, &res.Result.Reconnects, &res.Result.Time, &res.Result.Threads, &res.Result.QPS.ID,
+			&res.Result.QPS.Total, &res.Result.QPS.Reads, &res.Result.QPS.Writes, &res.Result.QPS.Other)
+		if err != nil {
+			return nil, err
+		}
+		macrodetails = append(macrodetails, res)
+	}
+	return macrodetails, nil
+}
+
```
harshit-gangal, line 125: I think this function is specific to the query as the scan have the columns list. I think this can be made inline.

go/server/handlers.go
frouioui: Fixed in d230e96ed7486779a71b7f916a9784bfca01b336

go/server/handlers.go
frouioui: Good idea. I have added a "debug" method to render error in c8a104b0a3ae8f25b745c268bde94f166121b41c

go/tools/macrobench/macrobench.go
frouioui: Definitely. Removed in 4c4a7d4fc9804e197b61707bfd76c0e6e2f4d517

go/tools/macrobench/results.go
frouioui: I have proposed a change in 4d0432599c6fa539f1d8d529cb3e33c9eb87401f.

go/server/handlers.go
```diff
@@ -20,21 +20,48 @@ package server
...
 
+func handleRenderErrors(c *gin.Context, err error) {
+	if err == nil {
+		return
+	}
+	log.Println(err.Error())
+	c.HTML(http.StatusOK, "error.tmpl", gin.H{
+		"title":    "Vitess benchmark - Error",
+		"url":      c.FullPath(),
+		"errorStr": err.Error(),
```
harshit-gangal, line 38: we should not give the internal error out. It should be some masked error message.

go/server/handlers.go
```diff
@@ -20,21 +20,48 @@ package server
...
+func (s *Server) homeHandler(c *gin.Context) {
+	oltpData, err := macrobench.GetResultsForLastDays("ee", "webhook", 31, s.dbClient)
+	if err != nil {
+		handleRenderErrors(c, err)
+		return
+	}
+
+	tpccData, err := macrobench.GetResultsForLastDays(macrobench.TPCC, "webhook", 31, s.dbClient)
+	if err != nil {
+		handleRenderErrors(c, err)
```
harshit-gangal, line 57: we should not return from first error. If we can show the second widget, we should show that atleast

frouioui: Resolved in 6383b39abc7496ef2ce05777b36bd323d14d6a72. Errors are logged serverside, but the page will be rendered. If some graphs can't be rendered, the text  _`The graph could not be rendered.`_ will be displayed instead.

go/server/handlers.go
frouioui: Thank you, that's a good suggestion. I went ahead and changed the way to render errors in the corresponding template. I have removed the `errorStr` field and used a static text instead. The error itself is still logged serverside. Fix was pushed in 3386b9122f9d8dd584c8af41f12e22204a09da22.


image

go/server/handlers.go
```diff
@@ -20,21 +20,45 @@ package server
...
 	c.HTML(http.StatusOK, "information.tmpl", gin.H{
 		"title": "Vitess benchmark",
 	})
 }
 
-func (s *Server) homeHanlder(c *gin.Context) {
+func (s *Server) homeHandler(c *gin.Context) {
+	oltpData, err := macrobench.GetResultsForLastDays(macrobench.OLTP, "webhook", 31, s.dbClient)
+	if err != nil {
+		log.Println(err.Error())
```
harshit-gangal, line 50: use log.Warn or log.Error

go/server/handlers.go
```diff
@@ -20,21 +20,45 @@ package server
...
+	}
+
+	tpccData, err := macrobench.GetResultsForLastDays(macrobench.TPCC, "webhook", 31, s.dbClient)
+	if err != nil {
+		log.Println(err.Error())
```
harshit-gangal, line 55: same as above.

harshit-gangal (approved):

frouioui:
@harshit-gangal - Rebased on master and added a new logger. That new logger is only in the `server` package for now. Though, I have made it possible to set the server's logger from a root package, for example before calling `srv.Run()` in the web CLI handler (`./go/cmd/web/web.go:34`).

PR #113, opened 2021-04-06
Fixed idempotence issue with Vitess Build and Ansible

frouioui:
## Description

The issue fixed in this pull request is linked to the idempotence step of `molecule test`. The `Molecule Test vitess_build` workflow started to fail due to the role `vitess_build` not being idempotent. More on that [here](https://github.com/vitessio/arewefastyet/actions/runs/718746242).

The error was the following:
```
--> Action: 'idempotence'
ERROR: 
PLAY [Converge] ****************************************************************
An error occurred during the test sequence action: 'idempotence'. Cleaning up.
```

## Fix

The fix is a new condition on the block of tasks where we fetch Vitess's repository and build its binaries. The condition checks if the repository already exists in the system. Doing so allows us to skip unnecessary steps and thus making the role idempotent.

### DIFF
--- a/ansible/roles/vitess_build/tasks/clean.yml
+++ b/ansible/roles/vitess_build/tasks/clean.yml
@@ -17,16 +17,13 @@
     systemctl stop cell@* || /bin/true
     systemctl stop vtgate@* || /bin/true
   ignore_errors: yes
-  changed_when: false
+  when: clean is defined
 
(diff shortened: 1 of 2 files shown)


harshit-gangal (approved):

harshit-gangal:
nice find.

PR #114, opened 2021-04-07
Integration of Webhooks API

### LINKED ISSUE #103: Integrate webhook in the new server
The webhook that was previously built in python needs to be copied to the new server in golang to run TPCC, OLTP, and Microbenchmarks.

frouioui:
## Description

Hello! The webhooks used to trigger builds from GitHub are implemented in Python code. This pull request brings them back to life by transferring them to the new Go code.

## Implementation

In order to achieve such transfer, a new `api.go` file has been added to the server's package. This file contains, for now, the route handling webhooks requests. To stay compatible with how we currently use Webhooks from GitHub, the interface and query did not change.

The following cURL command will trigger the new handler:

```bash
curl --location --request POST 'localhost:8080/webhook' --header 'Content-Type: application/json' --data-raw '{"ref": "4fd3140ba080db8f57e781317ed7e69233809891"}'
```

The `ref` in the body refers to the git SHA used.

The code used by the handler is fairly simple, it uses our main `Exec` interface. Reminder, `Exec` allows us to deal with the execution flow of any given task on any given infrastructure using a small number of methods. When receiving a webhook request, a new `Exec` is created, its UUID is returned to the caller, and a goroutine will start with the execution flow.

### So how to pass the configuration I want to Exec through webhook?

Exec's power is its configurability, though it is done through YAML configuration files or CLI files thanks to Cobra and Viper. Since passing the full configuration file through the webhook query's body can lead to a high level of maintenance (we need to set up all the configurations we want in GH Secrets, might have multiple configurations, risky secrets, ...), I have decided instead to store the configuration files that will be used by webhook's Exec directly on the remote machine (i.e the server). The configuration file used by default for each webhook's exec can be defined through a new CLI flag `--web-webhook-config`, this is mostly done to offer compatibility with the current way we're dealing with webhook on GitHub. The configuration file can also be defined in the query's body using the key `path_config`, this feature will be improved as we get rid of the "old way".

Once we have our configuration file for the execution, we create a new `*viper.Viper`, parse the file, and apply the configuration to our new `Exec`, in order to do that new methods `AddToViper` are added to multiple structs `AnsibleConfig`, `InfraConfig` and `Exec`.

## How to run

```bash
go build ./go/main.go

./main web --config ./config/config.yaml --web-webhook-config ./config/config_oltp_sharded.yaml
```

## Related issues

Resolves #103.


go/exec/exec.go
```diff
@@ -33,21 +44,78 @@ type Exec struct {
...
 	err := e.prepareDirectories()
 	if err != nil {
 		return err
 	}
 
+	err = e.Infra.Prepare()
+	if err != nil {
+		return err
+	}
+	e.prepared = true
```
systay, line 107: should this be checked so we don't accidentally prepare an Exec twice?

systay (approved):

go/exec/exec.go
frouioui: Good finding! Fixed in a80ed41e4e86a7c54a6e4e4a73a6028ab3482942.

PR #119, opened 2021-04-08
Add a macro benchmark command

### LINKED ISSUE #115: Add macro benchmarks as a separate tool/command
## Description

The current go code supports macro benchmarks through the use of `exec`, but we need to add support through a unique command such as what was done for `microbench run`. Doing so will allow us to follow the same pattern as the one we used for microbenchmarks. It will allow greater control over the reporting step.

frouioui:
## Description

This pull request adds a new command to the CLI. The command allows us to execute a macro benchmark independently. Just like what we already do for `microbench`, we will not have to **only** go through the `exec` command in order to start macro benchmarks.

## Why do we need this

I came up with four main reasons why we need to implement this change. _First_ of all, it offers more consistency in the codebase, we were already treating micro benchmark with a unique command, so why not macro benchmark. _Secondly_, it allows easy local debugging and increases reproducibility, there is no need to spin up the infrastructure and wait for it to be configured. _Thirdly_, it enables us to deal with results and errors in a better way, everything is executed and controlled within the Golang code, not third-party shell scripts. _And finally_, it brings us closer to a system with centralized configuration, we configure macro benchmark from our main `config.yaml` using Cobra.

Being a tenth skeptical at first about removing the responsibility of executing Sysbench to Ansible and passing it on to the Golang code, those four reasons are what drove me anyhow.

## The new command

The new command usage is:

```text
go run ./go/main.go  macrobench run --help
Usage:
  arewefastyet macrobench run [flags]

Flags:
      --db-database string                      Database to use.
      --db-host string                          Hostname of the database
      --db-password string                      Password to authenticate the database.
      --db-user string                          User used to connect to the database
  -h, --help                                    help for run
      --macrobench-git-ref string               Git SHA referring to the macro benchmark.
      --macrobench-skip-steps strings           Slice of sysbench steps to skip.
      --macrobench-source string                The source or origin of the macro benchmark trigger.
      --macrobench-sysbench-executable string   Path to the sysbench binary.
      --macrobench-type MacroBenchmarkType      Type of macro benchmark.
      --macrobench-working-directory string     Directory on which to execute sysbench.
      --macrobench-workload-path string         Path to the workload used by sysbench.
```

## Configurability

As I mentioned, this command offers increased configurability of the macro benchmarks. Instead of declaring variables for both OLTP and TPCC inside Ansible inventories, we define a single set of variables that can apply to both OLTP and TPCC in a single place. This set of variables will be read by the Golang code and transformed into a list of arguments that we pass to Sysbench. Below is an extract of a configuration file containing macro benchmarks configuration.

```yaml
#### MACRO BENCHMARKS CONFIGURATION

macrobench-sysbench-executable: /usr/local/bin/sysbench
macrobench-workload-path: oltp_read_only
macrobench-skip-steps:
macrobench-type: oltp
macrobench-source: local
macrobench-git-ref: head

## ALL
macrobench_all_mysql-db: main
macrobench_all_mysql-host: 127.0.0.1
macrobench_all_mysql-port: 13306
macrobench_all_db-ps-mode: disable
macrobench_all_db-driver: mysql
macrobench_all_luajit-cmd: "off"
macrobench_all_threads: 1
macrobench_all_auto-inc: "off"
macrobench_all_tables: 50
macrobench_all_table_size: 10
macrobench_all_range_selects: 0
macrobench_all_rand-type: uniform

## PREPARE
macrobench_prepare_time: 30
macrobench_prepare_report-interval: 10

## WARM UP
macrobench_warmup_time: 10
macrobench_warmup_report-interval: 10

## RUN
macrobench_run_time: 10
macrobench_run_report_json: true
macrobench_run_verbosity: 0
```

Variables are prefixed with `macrobench_` followed by the name of the sysbench step on which this variable applies. As we can see, the variables at the top of the file, those without a step prefix, are used to control the execution. They are not passed down to sysbench. Those can be configured through CLI flags, as well as a configuration file like that one.

The sysbench workload used by the macro benchmark is now configurable too through the `macrobench-workload-path` key. This takes the lua file that sysbench will use.

## Sysbench

We use forks of sysbench and sysbench-tpcc, they have been modified to generate JSON reports:
- https://github.com/planetscale/sysbench
- https://github.com/planetscale/sysbench-tpcc

## Testing

To test you'll need a sufficient configuration, with the following items:

> Here showing a config.yaml file used to execute TPCC macro benchmarks.

```yaml
equinix-token: <equinix token>
equinix-project-id: <equinix project id>
equinix-instance-type: m2.xlarge.x86

# database information is not mandatory, the command can ignore storing results
db-host: <your db hostname>
db-user: <your db user>
db-password: <your db password>
db-database: <your db name>

infra-path: ./infra/terraform/

ansible-inventory-files: macrobench_sharded_inventory.yml
ansible-playbook-files: macrobench.yml
ansible-root-directory: ./ansible/

#### MACRO BENCHMARKS CONFIGURATION

macrobench-sysbench-executable: /usr/local/bin/sysbench
macrobench-workload-path: /src/sysbench-tpcc/tpcc.lua
macrobench-skip-steps:
macrobench-type: tpcc
macrobench-source: local
macrobench-git-ref: head
macrobench-working-directory: /src/sysbench-tpcc

## ALL
macrobench_all_mysql-db: main
macrobench_all_mysql-host: 127.0.0.1
macrobench_all_mysql-port: 13306
macrobench_all_db-ps-mode: disable
macrobench_all_db-driver: mysql
macrobench_all_luajit-cmd: "off"
macrobench_all_threads: 1
macrobench_all_tables: 1
macrobench_all_scale: 1
macrobench_all_rand-type: uniform
macrobench_all_use_fk: 0
macrobench_all_mysql-ignore-errors: all

## PREPARE
macrobench_prepare_time: 5
macrobench_prepare_report-interval: 10

## WARM UP
macrobench_warmup_time: 5
macrobench_warmup_report-interval: 10

## RUN
macrobench_run_time: 10
macrobench_run_report_json: "yes"
macrobench_run_verbosity: 0
```

New inventories are added, they will permanently replace old ones in future PRs. Both `./ansible/macrobench_sharded_inventory.yml` and `./ansible/macrobench_unsharded_inventory.yml` can be used. Here is a sample of the configuration needed to execute those files:

```yaml
...
ansible-inventory-files: macrobench_sharded_inventory.yml
ansible-playbook-files: macrobench.yml
ansible-root-directory: ./ansible/
...
```

You can then execute the commands with two methods:

```sh
go build ./go/main.go

## Method 1: Execution from A to Z using Terraform and Ansible as interfaces
./main exec --config ./config/config.yaml

## Method 2: Local execution using the new command - Requires Sysbench to be installed
./main macrobench run --config ./config/config.yaml

```

## Cleaning and future work

Some housecleaning will need to be done in the future, some of the files located in `./ansible/` are redundant.

Future work includes setting up a cleaner configuration file and variables for sysbench's params. YAML supports arrays, I think that's something we can look at to make files look cleaner.

## Related issues

Resolves #115.

### DIFF
--- a/ansible/roles/sysbench/tasks/main.yml
+++ b/ansible/roles/sysbench/tasks/main.yml
@@ -81,7 +81,7 @@
   changed_when: false
 
 - name: Set up VTGate Loadbalancing
-  shell: iptables -t nat -A OUTPUT -p tcp --dport 3306 -m state --state NEW -m statistic --mode nth --every {{ ((gateways | length) - index) }} --packet 0 -j DNAT --to-destination {{ gateway }}
+  shell: iptables -t nat -A OUTPUT -p tcp --dport 13306 -m state --state NEW -m statistic --mode nth --every {{ ((gateways | length) - index) }} --packet 0 -j DNAT --to-destination {{ gateway }}
   with_items: '{{ gateways }}'
   changed_when: false
   loop_control:
--- a/go/cmd/macrobench/run.go
+++ b/go/cmd/macrobench/run.go
@@ -25,14 +25,14 @@ import (
 )
 
 func run() *cobra.Command {
-	mabcfg := macrobench.MacroBenchConfig{
+	mabcfg := macrobench.Config{
 		DatabaseConfig: &mysql.ConfigDB{},
 	}
 
 	cmd := &cobra.Command{
 		Use: "run",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := macrobench.MacroBench(mabcfg)
+			err := macrobench.Run(mabcfg)
 			if err != nil {
 				return err
 			}
--- a/go/cmd/macrobench/run.go
+++ b/go/cmd/macrobench/run.go
@@ -20,13 +20,13 @@ package macrobench
 
 import "github.com/spf13/cobra"
 
-func MacroBenchCmd() *cobra.Command {
+func run() *cobra.Command {
 	cmd := &cobra.Command{
-		Use: "macrobench <command>",
-		Aliases: []string{"mab"},
+		Use: "run",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// todo: run the macrobench
+			return nil
+		},
 	}
-
-	cmd.AddCommand(run())
-
 	return cmd
 }
--- a/go/cmd/macrobench/run.go
+++ b/go/cmd/macrobench/run.go
@@ -32,11 +32,7 @@ func run() *cobra.Command {
 	cmd := &cobra.Command{
 		Use: "run",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := macrobench.Run(mabcfg)
-			if err != nil {
-				return err
-			}
-			return nil
+			return macrobench.Run(mabcfg)
 		},
 	}
 	mabcfg.AddToCommand(cmd)
--- a/go/tools/macrobench/config.go
+++ b/go/tools/macrobench/config.go
@@ -107,9 +107,9 @@ func (mabcfg *MacroBenchConfig) parseIntoMap(prefix string) {
 	}
 }
 
-// InsertBenchmarkToSQL will insert a new row in the benchmark table based on
+// insertBenchmarkToSQL will insert a new row in the benchmark table based on
 // the given MacroBenchConfig. The newly created row's unique ID is returned.
-func (mabcfg MacroBenchConfig) InsertBenchmarkToSQL(client *mysql.Client) (newMacroBenchmarkID int, err error) {
+func (mabcfg MacroBenchConfig) insertBenchmarkToSQL(client *mysql.Client) (newMacroBenchmarkID int, err error) {
 	if client == nil {
 		return 0, errors.New(mysql.ErrorClientConnectionNotInitialized)
 	}
--- a/go/tools/macrobench/config.go
+++ b/go/tools/macrobench/config.go
@@ -21,24 +21,12 @@ package macrobench
 import (
 	"github.com/spf13/cobra"
 	"github.com/vitessio/arewefastyet/go/mysql"
-	"github.com/vitessio/arewefastyet/go/tools/macrobench"
 )
 
-func run() *cobra.Command {
-	mabcfg := macrobench.MacroBenchConfig{
-		DatabaseConfig: &mysql.ConfigDB{},
-	}
+type MacroBenchConfig struct {
+	DatabaseConfig *mysql.ConfigDB
+}
 
-	cmd := &cobra.Command{
-		Use: "run",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			err := macrobench.MacroBench(mabcfg)
-			if err != nil {
-				return err
-			}
-			return nil
-		},
-	}
-	mabcfg.AddToCommand(cmd)
-	return cmd
+func (mabcfg *MacroBenchConfig) AddToCommand(cmd *cobra.Command)  {
+	mabcfg.DatabaseConfig.AddToCommand(cmd)
 }
--- a/go/tools/macrobench/config.go
+++ b/go/tools/macrobench/config.go
@@ -107,9 +107,9 @@ func (mabcfg *MacroBenchConfig) parseIntoMap(prefix string) {
 	}
 }
 
-// RegisterNewBenchmarkToMySQL will insert a new row in the benchmark table based on
+// InsertBenchmarkToSQL will insert a new row in the benchmark table based on
 // the given MacroBenchConfig. The newly created row's unique ID is returned.
-func (mabcfg MacroBenchConfig) RegisterNewBenchmarkToMySQL(client *mysql.Client) (newMacroBenchmarkID int, err error) {
+func (mabcfg MacroBenchConfig) InsertBenchmarkToSQL(client *mysql.Client) (newMacroBenchmarkID int, err error) {
 	if client == nil {
 		return 0, errors.New(mysql.ErrorClientConnectionNotInitialized)
 	}
--- a/go/tools/macrobench/config.go
+++ b/go/tools/macrobench/config.go
@@ -33,6 +33,7 @@ type MacroBenchConfig struct {
 	DatabaseConfig *mysql.ConfigDB
 	M              map[string]string
 	SkipSteps      []string
+	Type           MacroBenchmarkType
 }
 
 const (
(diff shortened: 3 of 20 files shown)


harshit-gangal (commented):
overall looks good.

few nits
- add flag description where missing
- export the functions which are used outside of the package, others you can make it internal.

ansible/roles/sysbench/tasks/main.yml
```diff
@@ -81,7 +81,7 @@
   changed_when: false
 
 - name: Set up VTGate Loadbalancing
-  shell: iptables -t nat -A OUTPUT -p tcp --dport 3306 -m state --state NEW -m statistic --mode nth --every {{ ((gateways | length) - index) }} --packet 0 -j DNAT --to-destination {{ gateway }}
+  shell: iptables -t nat -A OUTPUT -p tcp --dport 13306 -m state --state NEW -m statistic --mode nth --every {{ ((gateways | length) - index) }} --packet 0 -j DNAT --to-destination {{ gateway }}
```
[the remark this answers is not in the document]
frouioui: Yes. I changed the default port that points to vtgate's load balancer (from `3306` to `13306`) because all tcp packets with port `3306` were routed towards one of the vtgates. However, we not only want to communicate with vtgates,  we also want to communicate with other MySQL hosts, for instance: the SQL database used to store results.

This is why the configuration file contains the variable shown below, it allows us to declare the port of our vtgate load balancer.

```yaml
macrobench_all_mysql-port: 13306
```

frouioui:
> overall looks good.
> 
> few nits
> 
> * add flag description where missing
> * export the functions which are used outside of the package, others you can make it internal.

Fixed those nits in 2a28dcd52b07d6ff8d342f2fa1acc435e671b294, 82c480402f53d1a0f92730c34d24f630ef3894fd and a2b3e5576fa57acbdb93921dc75d1ad23f4fb934.

go/tools/macrobench/config.go
```diff
@@ -0,0 +1,123 @@
...
+	"errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/vitessio/arewefastyet/go/mysql"
+	"strings"
+)
+
+// MacroBenchConfig defines a configuration used to execute macro benchmark.
+// For instance, the MacroBench method uses MacroBenchConfig.
+type MacroBenchConfig struct {
```
harshit-gangal, line 31: nit: code styling thing, the name starts with package name. This can be named just Config

go/tools/macrobench/macrobench.go
```diff
@@ -0,0 +1,138 @@
...
+			resStr = out
+		}
+	}
+
+	// Parse results
+	var results []MacroBenchmarkResult
+	err = json.Unmarshal(resStr, &results)
+	if err != nil {
+		return fmt.Errorf("unmarshal results: %+v\n", err)
+	} else if len(results) < 0 {
```
harshit-gangal, line 126: nit: else is not required if you return from the if condition.

go/tools/macrobench/macrobench.go
```diff
@@ -0,0 +1,138 @@
...
+		return errors.New(ErrorNoSysBenchResult)
+	}
```
harshit-gangal, line 128: i think you wanted to check for `len(results) == 0`

go/tools/macrobench/macrobench.go
```diff
@@ -0,0 +1,138 @@
...
+	var err error
+	var sqlClient *mysql.Client
+
+	if mabcfg.DatabaseConfig != nil && mabcfg.DatabaseConfig.IsValid() {
+		sqlClient, err = mysql.New(*mabcfg.DatabaseConfig)
+		if err != nil {
+			return err
+		}
+		defer sqlClient.Close()
+	}
```
harshit-gangal, line 87: I think instead of this. If the condition fails then you should throw an error and then you would not be needed to check later for `sqlclient != nil`

go/tools/macrobench/results.go
```diff
@@ -107,3 +109,30 @@ func GetResultsForLastDays(macroType MacroBenchmarkType, source string, lastDays
...
+// insertToMySQL inserts the given MacroBenchmarkResult to MySQL using a *mysql.Client.
+// The MacroBenchmarkResults gets added in one of macrobenchmark's children tables.
+// Depending on the MacroBenchmarkType, the insert will be routed to a specific children table.
+// The children table QPS is also inserted.
+func (mbr *MacroBenchmarkResult) insertToMySQL(benchmarkType MacroBenchmarkType, macrobenchmarkID int, client *mysql.Client) error {
+	if client == nil {
+		return errors.New(mysql.ErrorClientConnectionNotInitialized)
+	} else if benchmarkType == "" {
+		return errors.New(IncorrectMacroBenchmarkType)
+	}
```
harshit-gangal, line 122: nit: change remove else as the if condition has return.

go/tools/macrobench/config.go
frouioui: Changed the name in 4666fa63a55c35c85a879faa2ea2879cd7a6ae09. Following the same thought, I changed `MacroBench` method to `Run`.

go/tools/macrobench/macrobench.go
frouioui: Good catch, fixed.

go/tools/macrobench/macrobench.go
frouioui: I do not want to restrict the use of SQL. The user must be able to run a macro benchmark without necessarily saving its results to a database. If `mabcfg.DatabaseConfig == nil` then `sqlclient` stays `nil` and no error is thrown.

go/tools/macrobench/results.go
frouioui: Sounds good, fixed them both in 0f13367.

go/cmd/macrobench/run.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use: "run",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := macrobench.Run(mabcfg)
+			if err != nil {
+				return err
+			}
+			return nil
```
GuptaManan100, line 39: nit: return err

GuptaManan100 (approved):
✨

go/cmd/macrobench/run.go
frouioui: Resolved in 20796f5.

PR #125, opened 2021-04-13
New parent table for executions

### LINKED ISSUE #117: Add a parent table to macro and micro benchmarks
## Description

We need a parent table to contain information on each execution. Each row will be composed of a single execution's informations, such as:

- The date and time it started
- The date and time it finished
- Where it was triggered
- The type of underlying execution (is it a microbenchmark, a macro, OLTP, TPCC?)
- The status of the execution (creating, started, finished, failed, skipped, ...)

To do this, the current naming and schema of the database will need to be changed. For example, the `benchmark` table should probably be renamed into `macrobenchmark`.

frouioui:
## Description

> WIP

## Related issues

Resolves #117.

### DIFF
--- a/sql/vitess-benchmark.sql
+++ b/sql/vitess-benchmark.sql
@@ -1,4 +1,16 @@
-CREATE TABLE `OLTP` (
+CREATE DATABASE IF NOT EXISTS benchmark;
+
+USE benchmark;
+
+CREATE TABLE IF NOT EXISTS macrobenchmark (
+  `test_no` int(11) NOT NULL AUTO_INCREMENT,
+  `commit` varchar(100) DEFAULT NULL,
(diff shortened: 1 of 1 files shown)


go/exec/status.go
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package exec
+
+type (
+	status string
+)
+
+var (
+	statusCreated  = status("created")
```
systay, line 26: shouldn't these be consts?

systay (approved):
Nicely commented. Makes it much easier to understand

go/exec/status.go
frouioui: The use of variables is indeed not required. I will change it

frouioui:
Commits 032e6c8..8cc391d introduce a new MySQL migration workflow. New migrations are added to the `./sql/migration` directory, the `.sql` files are prefixed by the migration ID (`000`, `001`, `002`, ...).

PR #126, opened 2021-04-14
Search and Compare in Go server

### LINKED ISSUE #123: Integrate search and compare features in the Go server
## Description

While we moved most of the Python's server features to the new Golang code, the `Search or Compare` pannel is still unavailable. It should be reintegrated using the new interfaces.

Comparing two SHAs should return the macro and micro benchmarks comparison.

frouioui:
## Description

Integration of two pages Search and Compare, which were integrated into a single page on the old Python code. They both enable visualization of benchmark results. These pages will serve as the founding block of release reports. Comparing two releases (two commit SHA) will be done through the compare page. The goal is to humanize the macro benchmarks results by adding charts, difference %, tables. Micro benchmark results will be rendered through the existing table.

## Search

The search page allows us to see results for a given commit SHA.

image

## Compare

The comparison page contains graphs for each type of macro benchmark like the one illustrated below.

image

## Refactored

- Refactored the `macrobench` package with a cleaner naming convention. Structs and methods that had  `MacroBench` (or similar) in their names were changed and shortened. Example: `MacroBenchmarkType` becomes `Type`.

## Related issues

Resolves #123.


go/tools/git/git.go
```diff
@@ -33,4 +33,13 @@ func GetCommitHash(repoDir string) (hash string, err error) {
 
 	hash = ref.Hash().String()
 	return hash, nil
+}
+
+// ShortenSHA will return the first 7 characters of a SHA.
+// If the given SHA is too short, it will be returned untouched.
+func ShortenSHA(sha string) string {
+	if len(sha) > 7 {
```
systay, line 41: is 7 enough for us?

frouioui: `ShortenSHA` will return the commit SHA like `git` does, instead of `69eaf6578535a1b053ae9f84bc2e2d7530738c8f` they become `69eaf65`, 7 is the length of the latter string

frouioui: In the web UI we will use both long and short strings. The shorter string will be used to render information and avoid having a long text. The longer string will be used by hyperlinks and will redirect users to vitessio/vitess.

go/server/handlers.go
```diff
@@ -61,9 +62,104 @@ func (s *Server) homeHandler(c *gin.Context) {
...
+	if search == "" {
+		c.HTML(http.StatusOK, "search.tmpl", gin.H{
+			"title": "Vitess benchmark",
+		})
+		return
+	}
+
+	// Get macro benchmarks from all the different types
+	macros := map[macrobench.Type]interface{}{}
+	for _, mtype := range macrobench.Types {
```
systay, line 141: this is almost the same for loop that we have on line 92. wdyt about extracting a method that can be used from both places?

frouioui: Good idea, I have extracted the method via 20c6de4

systay (approved):

PR #129, opened 2021-04-15
Changed default vitess package for microbenchmark

frouioui:
## Description

Changed the default Vitess package used by microbenchmark from `./go/vt/sqlparser/...` to `./go/vt/...`.


GuptaManan100 (approved):
🚀

PR #142, opened 2021-05-03
Drop test_no column in microbenchmark

frouioui:
## Description

Inserting micro-benchmarks results to MySQL using the schema defined by `001` was producing the following error:

```
vttablet: rpc error: code = Unknown desc = Field 'test_no' doesn't have a default value (errno 1364) (sqlstate HY000) (CallerID: vitess): Sql: "insert into microbenchmark(exec_uuid, pkg_name, `name`, git_ref) values (NULLIF(:v1, :vtg1), :v2, :v3, :v4)"
```

In fact, migration `001` was missing a `DROP COLUMN` statement, even though the complete schema (in `./sql/vitess-benchmark.sql`) is correct. To fix this issue, the missing statement is appended to the migration `001` file.


GuptaManan100 (approved):
✨

PR #167, opened 2021-05-11
Add notification in case of regression

### LINKED ISSUE #165: Add notification in case of regression
## Description

Cron jobs must be able to send a slack message when there is a regression in benchmark results.

frouioui:
## Description

This pull request enable sending Slack message in the event of a benchmark regression. After each cron job execution, a new step is added `SendNotification`. This step will compute the difference between the current job and the previous job, and determine if a notification needs to be sent. If so, a message will be sent to Slack using the provided configuration.

To achieve this, the `execution` table had to be updated, a new column `type` stores what type of execution we are dealing with, `micro`, `oltp`, or `tpcc`. Migration `003` is added for this purpose. A typography issue was fixed in `001` as well.

The Slack messages are built using Block Kit, the format looks as followed:

**Micro benchmark:**
```txt
Observed a regression.
Comparing: recent commit {{New Commit}} with old commit {Old Commit}.
Benchmark UUIDs, recent: {New Exec UUID} old: {Old Exec UUID}.
- {Benchmark Package}/{Benchmark Name} decreased by {Decrease Rate}%
```

**Macro benchmark:**
```txt
Observed a regression.
Comparing: recent commit {{New Commit}} with old commit {Old Commit}.
Benchmark UUIDs, recent: {New Exec UUID} old: {Old Exec UUID}.
- Total CPU time increased by {Increase Rate}%
- vttablet CPU time increased by {Increase Rate}%
- vtgate CPU time increased by {Increase Rate}%
- TPS decreased by {Decrease Rate}%
- QPS decreased by {Decrease Rate}%
- Latency increased by {Increase Rate}%
```

## Related issues

Fixes #165


go/exec/exec.go
```diff
@@ -206,6 +217,78 @@ func (e *Exec) Execute() (err error) {
 	return nil
 }
 
+func (e Exec) SendNotification() error {
```
GuptaManan100, line 220: nit: Rename to `SendNotificationForRegression`. `SendNotification` seems like a generic function name which can be used to send any notification

go/tools/microbench/compare.go
```diff
@@ -43,3 +44,18 @@ func CompareMicroBenchmarks(dbClient *mysql.Client, reference string, compare st
...
+
+// Regression returns a string containing the reason of the regression of the given MicroBenchmarkComparisonArray,
+// if no regression was evaluated, the reason will be an empty string.
+// The format of a single benchmark regression's reason is like this:
+//
+// "- {pkg name}/{benchmark name} decreased by {decrease percentage}%\n"
+//
+func (microsMatrix MicroBenchmarkComparisonArray) Regression() (reason string) {
+	for _, micro := range microsMatrix {
+		if micro.CurrLastDiff < 0.90 {
```
GuptaManan100, line 56: For now this is fine, we can change these numbers for regression if there are too many or too little slack notifications

GuptaManan100 (approved):
🎊

go/exec/exec.go
frouioui: Fixed via 6caea90

Akilan1999 (approved):

PR #187, opened 2021-05-18
Compare MacroBenchmarks of releases like Microbenchmarks

### LINKED ISSUE #186: Compare MacroBenchmarks of releases like Microbenchmarks

GuptaManan100:
## Description

This PR adds a new section to the website which allows comparison of macrobenchmarks on releases just like the macrobenchmarks.
Screenshot 2021-05-18 at 12 33 09 PM

## Linked Issues
Resolves #186

### DIFF
--- a/go/server/templates/macrobench.tmpl
+++ b/go/server/templates/macrobench.tmpl
@@ -178,6 +178,86 @@ limitations under the License.
                 window.open("/macrobench?ltag="+$('#leftTagDropdownMenuButton').text()+"&rtag="+$('#rightTagDropdownMenuButton').text(), '_self');
             })
         });
+
+        function macroCmpBarCharts(type, macro) {
+            for (const macroKey in macro) {
+                var compareMacros = macro[macroKey]["Compare"]["Result"]
+                var referenceMacros = macro[macroKey]["Reference"]["Result"]
+
+                var labels = []
+                var dataCmp = []
+                var dataRef = []
+                var wants = ["total", "reads", "writes", "other"]
+                for (var x in referenceMacros["qps"]) {
+                    if (wants.includes(x)) {
+                        labels.push(x)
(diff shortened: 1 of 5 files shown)


go/server/templates/macrobench.tmpl
```diff
@@ -0,0 +1,267 @@
...
+                        labels.push(x)
+                        dataCmp.push(compareMacros["qps"][x])
+                        dataRef.push(referenceMacros["qps"][x])
+                    }
+                }
+                const data = {
+                    labels: labels,
+                    datasets: [
+                        {
+                            label: '{{.compare.short}}',
```
frouioui, line 202: This label does not work properly, I think you should use `.leftTag` instead.

go/server/templates/macrobench.tmpl
```diff
@@ -0,0 +1,267 @@
...
+                            ],
+                            borderColor: [
+                                'rgba(255, 99, 132, 1)',
+                                'rgba(255, 99, 132, 1)',
+                                'rgba(255, 99, 132, 1)',
+                                'rgba(255, 99, 132, 1)',
+                            ],
+                        },
+                        {
+                            label: '{{.reference.short}}',
```
frouioui, line 219: Same as above, the variable is not defined, you should use `.rightTag` here.

frouioui (commented):
Awesome change, just left two small comments regarding the charts' labels.

Additionally, on the new `/macrobench` endpoint, the columns that render a tag's results are inversed. When choosing `9.0.0` for the left, and `10.0.0` for the right, the output table shows `9.0.0` on the right and `10.0.0` on the left. When compared with the microbenchmark table, we keep left on the left and right on the right. Can we get the same behavior on the macrobenchmarks page?

frouioui (approved):
Looks good to me!! Thank you 🚀

PR #192, opened 2021-05-18
Adds check for master cron job to check if there are existing runs for the same sha

### LINKED ISSUE #190: Master cron job does not check if there are existing runs for the same sha
**Describe the bug**
Master cron job does not check if there are existing runs for the same sha. This would happen when there were no commits on Vitess on a given day. In this case, we would not want to run the cron jobs for the same sha, but for now, there are no checks against this.

GuptaManan100:
## Description
Master cron job does not check if there are existing runs for the same sha. This would happen when there were no commits on Vitess on a given day. In this case, we would not want to run the cron jobs for the same sha. This PR adds a check for it.

## Related Issues
Fixes #190

### DIFF
--- a/go/server/cron.go
+++ b/go/server/cron.go
@@ -45,10 +45,10 @@ func (s *Server) createNewCron() error {
 }
 
 func (s *Server) cronMasterHandler() {
-	configs := []string{
-		s.microbenchConfigPath,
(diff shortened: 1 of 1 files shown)


frouioui (approved):
Perfect

PR #197, opened 2021-05-18
Addition of link for comparison in slack message

### LINKED ISSUE #193: Addition of link for comparison in slack message
## Description

In the regression slack message that we send, instead of just having the results, we would also want to add the link to the website where the comparison can be graphically seen.

GuptaManan100:
## Description

In the regression slack message that we send, instead of just having the results, we would also want to add the link to the website where the comparison can be graphically seen. This PR adds this functionality

## Linked Issues
Fixes #193


frouioui (approved):
Looks great. That'll be very useful!

PR #199, opened 2021-05-18
Restart cron job after a failure

### LINKED ISSUE #189: Restart cron job after a failure
## Description

Automatically restarting a cron job after their failure should be allowed, a retry argument can define the maximum number of retries before giving up. Restarting a job must create a new execution, not override the failed one.

frouioui:
## Description

This pull request adds a mechanism to retry executions triggered by cron jobs when they fail. The flag `web-cron-nb-retry` defines the number of retries that are allowed per execution.

## Related issues

Resolves #189.


GuptaManan100 (approved):
🕶️

PR #205, opened 2021-05-19
Patched release tags ordering

frouioui:
## Description

This pull request fixes the way we order git release tags.

The order before:

- 10.0.0
- 10.0.0-rc1
- 10.0.1
- 7.0.0
- 7.0.1
- ...
- 9.0.1
- master

The order now:

- master
- 10.0.1
- 10.0.0
- 10.0.0-rc1
- 9.0.1
- ...
- 7.0.1
- 7.0.0

PR #206, opened 2021-05-19
Addition of unit test for git releases ordering

frouioui:
## Description

Unit test for the proper ordering of git release tags which was introduced by #205.

### DIFF
--- a/go/tools/git/git_test.go
+++ b/go/tools/git/git_test.go
@@ -65,12 +65,12 @@ func TestGetAllVitessReleaseCommitHashOrdering(t *testing.T) {
 
 	// ordered releases from v10.0.1 to v9.0.0-rc1
 	tc := []*Release{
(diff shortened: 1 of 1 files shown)


frouioui:
> This test will start failing as soon as we have one more release, which would defeat the purpose of automation. Could you instead check that there should be a subset of the output which contains these release tags

Correct, will do!

GuptaManan100 (commented):
The test is failing currently. You need to include the numbers in the expectation too. That is, 
```go
{
  Name: "10.0.1", 
  CommitHash: "f7304cd1893accfefee0525910098a8e0e68deec",
  Number := []int{10,0,1},
},
```

frouioui:
Saw that, #207 added new fields. I'll push a fix

GuptaManan100 (approved):

PR #217, opened 2021-05-20
Golang memory allocs metrics support

### LINKED ISSUE #216: Add memory consumptions to exec metrics

frouioui:
## Description

This pull request provides initial support for golang's memory metrics. The metrics we're fetching and analyzing are `go_memstats_alloc_bytes_total`. In the future, it would be preferable to have a very flexible way of adding and modifying the golang metrics we want to fetch & compute.

The metrics are fetched in the execution metrics package and appended to macrobenchmark's comparison and details. The metric is stored for both VTGate and VTTablet, the sum of the two is also stored.

## Related issue(s)

Resolves #216.

### DIFF
--- a/go/exec/metrics/metrics.go
+++ b/go/exec/metrics/metrics.go
@@ -24,6 +24,18 @@ import (
 	awftmath "github.com/vitessio/arewefastyet/go/tools/math"
 )
 
+const (
+	cpuSecondsPerComponent = `from(bucket:"%s")
+			|> range(start: %s, stop: %s)
+			|> filter(fn:(r) => r._measurement == "process_cpu_seconds_total" and r.exec_uuid == "%s" and r.component == "%s")
+			|> max()`
+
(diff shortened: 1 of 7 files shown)


go/exec/metrics/metrics.go
```diff
@@ -53,37 +74,40 @@ type (
...
+			return ExecutionMetrics{}, err
+		}
+		execMetrics.TotalComponentsMemStatsAllocBytes += execMetrics.ComponentsMemStatsAllocBytes[component]
 	}
 	return execMetrics, nil
 }
 
 // getCPUTimeForComponent return the CPU Time taken by a component (vtgate, vttablet, ...)
 // for the given execUUID. The range of the select is defined through the start and end
 // arguments, to select the whole span one can use: "start:0, end:now()".
```
GuptaManan100, line 99: Change the comment here

GuptaManan100 (approved):
😸

PR #222, opened 2021-05-21
Planner Version Support in Macrobenchmarks

### LINKED ISSUE #214: v3 vs gen4
We should compare the macrobenchmark runs using the v3 and the gen4 planner, so we can see when gen4 start making a difference

GuptaManan100:
## Description

This PR adds a new flag to the execution of macrobenchmarks which allows us to change the planner version that vtgate is using - (V3, Gen4Fallback). It also updates the cron jobs to run for both versions for planner.
Finally, it adds additional tags to the equinix servers that are running.

## Linked Issues

Fixes #214

### DIFF
--- a/ansible/roles/macrobench/tasks/main.yml
+++ b/ansible/roles/macrobench/tasks/main.yml
@@ -42,6 +42,6 @@
 
 - name: Run macrobenchmarks
   shell: |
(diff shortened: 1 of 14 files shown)


frouioui (approved):
🚀

PR #227, opened 2021-05-21
Switch between vtgate planner V3 and Gen4

frouioui:
## Description

This pull request adds a switch to change which vtgate planner we want to use to query results on the website. The default is to use V3 planner, but can be switched to `Gen4Fallback` by turning the switch on.


GuptaManan100 (approved):
🍭

PR #228, opened 2021-05-21
Fixing compareSafeComponentMap when left map is empty

frouioui:
## Description

Fixing a use case where `metrics.compareSafeComponentMap`'s left map is empty and caused the output to be empty as well.


GuptaManan100 (commented):
🍝

GuptaManan100 (approved):

PR #233, opened 2021-05-24
Set cookie default instead of warning

### LINKED ISSUE #232: Cookie Error
The website logs show cookie reading errors.

GuptaManan100:
## Description

The first time the website is accessed, the `vtgatePlanner` cookie is not set. For now, we would throw a warning when this happened. The behavior has been changed to just return the default.

## Related Issues
Fixes #232

### DIFF
--- a/go/server/handlers.go
+++ b/go/server/handlers.go
@@ -51,14 +51,14 @@ func getPlannerVersion(c *gin.Context) macrobench.PlannerVersion {
 	planner := macrobench.V3Planner
 	plannerStr, err := c.Cookie("vtgatePlanner")
 	if err != nil {
-		slog.Warn(err.Error())
+		// cookie is not set, then use the default
+		return planner
 	}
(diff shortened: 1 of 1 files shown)


go/server/handlers.go
```diff
@@ -51,15 +51,15 @@ func getPlannerVersion(c *gin.Context) macrobench.PlannerVersion {
 	planner := macrobench.V3Planner
 	plannerStr, err := c.Cookie("vtgatePlanner")
 	if err != nil {
-		slog.Warn(err.Error())
+		// cookie is not set, then use the default
+		return planner
 	}
```
frouioui, line 56: The whole statement can be removed

frouioui (commented):
Just a small nit 👍🏻

go/server/handlers.go
GuptaManan100: Yeah I know, but I thought making it explicit will be more readable

frouioui (approved):

PR #235, opened 2021-05-25
Changing InfluxDB default settings for max-series-per-database 

frouioui:
## Description

Changing the `max-series-per-database` configuration to `0` (unlimited). The default value (1000000) got reached pretty fast and was causing an outage, the default retention policy will handle old results.

The error we got is written below and documented [here](https://docs.influxdata.com/influxdb/v1.8/troubleshooting/errors/#error-max-series-per-database-exceeded--).

```
error: max series per database exceeded: < >
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -195,3 +195,4 @@ override.tf
 override.tf.json
 *_override.tf
 *_override.tf.json
+arewefastyetcli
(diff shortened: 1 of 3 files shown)


GuptaManan100 (approved):
🧠

PR #237, opened 2021-05-25
Execute With Timeout and enhanced clean up

frouioui:
## Description

This pull request enhances the clean-up and execution workflow for cron jobs. A new method is added to exec: `ExecuteWithTimeout(timeout time.Duration)`, this method wraps the original `Execute` method with a timeout that will stop the execution if `timeout` is reached. The clean-up step is enhanced by adding the [flag](https://www.terraform.io/docs/cli/commands/init.html#lock-timeout-lt-duration-gt-) `-lock-timeout=<duration>` to Terraform's destroy command. This flag allows us to wait for terraform state's lock to be unlocked before attempting to destroy the resource, this tweak is useful in the event where we have to destroy our Equinix Metal instance even though its provision is not complete.

### DIFF
--- a/go/exec/exec.go
+++ b/go/exec/exec.go
@@ -196,7 +196,10 @@ func (e *Exec) Prepare() error {
 }
 
 // ExecuteWithTimeout will call execution's Execute method with the given timeout.
-func (e Exec) ExecuteWithTimeout(timeout time.Duration) error {
+func (e Exec) ExecuteWithTimeout(timeout time.Duration) (err error) {
+	defer func() {
+		e.handleStepEnd(err)
+	}()
 	errs := make(chan error)
 
 	go func() {
@@ -204,10 +207,11 @@ func (e Exec) ExecuteWithTimeout(timeout time.Duration) error {
 	}()
 
 	select {
-	case err := <-errs:
-		return err
+	case err = <-errs:
+		return
 	case <-time.After(timeout):
-		return errors.New(ErrorExecutionTimeout)
+		err = errors.New(ErrorExecutionTimeout)
+		return
 	}
 }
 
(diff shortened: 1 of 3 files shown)


go/exec/exec.go
```diff
@@ -193,6 +195,23 @@ func (e *Exec) Prepare() error {
 	return nil
 }
 
+// ExecuteWithTimeout will call execution's Execute method with the given timeout.
+// Note that a timeout too small will result in
```
GuptaManan100, line 199: incomplete comment

go/infra/equinix/equinix.go
```diff
@@ -87,7 +87,7 @@ func (e *Equinix) CleanUp() error {
 	for _, varValue := range e.TerraformVarArray() {
 		destroyOpts = append(destroyOpts, varValue)
 	}
-
+	destroyOpts = append(destroyOpts, tfexec.LockTimeout("120s"))
```
GuptaManan100, line 90: what is 120s over here

go/exec/exec.go
```diff
@@ -193,6 +195,23 @@ func (e *Exec) Prepare() error {
...
+
+	go func() {
+		errs <- e.Execute()
+	}()
+
+	select {
+	case err := <-errs:
+		return err
+	case <-time.After(timeout):
+		return errors.New(ErrorExecutionTimeout)
```
GuptaManan100, line 211: should we not cleanup e over here. Even if it timedout, the server needs to be closed

frouioui: We let the caller of `ExecuteWithTimeout` deal with the error, as we do with `Execute`

go/exec/exec.go
frouioui: Resolved via 2e32ccf

GuptaManan100 (approved):
:octocat:

PR #242, opened 2021-05-28
Fix Cron Jobs

### LINKED ISSUE #241: Fix Cron Jobs
## Description

There are some bugs and feature requests that need to be added to the cron jobs

- [BUG]: Currently the benchmarks are running for the same configuration multiple times. This issue arises because we are only testing for `finished` benchmarks while spawning new ones and not the ones which are in execution.
- [Feature]: Only send slack notifications for comparisons that are new and do not repeat the same message that we have already sent earlier.
- [BUG]: While adding Gen4 capability, we introduced a bug because we aren't checking the Vitess version and `planner-version` flag was introduced in 10.0.0. This means that all macro benchmark runs fail for older versions!

frouioui:
## Description

This pull request resolves various bugs and a feature request, all of them are detailed in #241.

- VTGate's `-planner_version` flag is used only when the git ref used is recent enough and includes the flag, meaning that git ref below V10.0.0 will not be benchmarked using the Gen4 planner.

- Only send slack notifications for comparisons that are new and do not repeat the same message that we have already sent earlier.

- Currently, the benchmarks are running for the same configuration multiple times. This issue arises because we are only testing `ifExists` for `finished` benchmarks while spawning new ones and not the ones which are in execution.

## Related issue(s)

Resolves #241.

### DIFF
--- a/ansible/roles/macrobench/tasks/main.yml
+++ b/ansible/roles/macrobench/tasks/main.yml
@@ -42,6 +42,6 @@
 
 - name: Run macrobenchmarks
   shell: |
-    arewefastyetcli macrobench run --config /tmp/config.yaml --macrobench-git-ref {{ vitess_git_version }} --macrobench-exec-uuid {{ arewefastyet_exec_uuid }} --macrobench-source {{ arewefastyet_source }} --macrobench-vtgate-planner-version {{ planner_version }}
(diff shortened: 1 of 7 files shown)


GuptaManan100 (approved):
💕

PR #247, opened 2021-06-09
Tooltip on gen4 toggle to ease navigation

frouioui:
Addition of a tooltip around the Gen4 toggle to ease understandability and navigation.


GuptaManan100 (approved):
LGTM 💟

PR #254, opened 2021-06-21
Changed the ordering of columns in webserver result tables

frouioui:
## Description

Changes the default order of the columns in the UI's result tables. The latest commit (on `main`) is now on the right-hand side of the table, allowing the `Improved by %` column to reflect the changes from the `latest release` to `main`.

**Before:**
image

**After:**
image


GuptaManan100 (approved):
LGTM! 💕

PR #257, opened 2021-06-22
Fixing v3vsGen4 macro benchmark results graphs order

frouioui:
## Description

The graphs used to render the results of v3vsGen4 macrobenchmarks had their columns inversed, and thus not matching the data they were being labeled for. This pull request fixes that.

**Before:**
image

**Now:**
image


GuptaManan100 (approved):
LGTM 🏎️

PR #259, opened 2021-06-28
Addition of a script to migrate execution metrics from influxDB to SQL

frouioui:
## Description

This pull request adds a script that, for all execution, fetches its metrics from influxDB and stores them to SQL if they are not already present.


GuptaManan100 (approved):
LGTM 🃏

PR #260, opened 2021-06-29
Fixed improved by % for the macrobenchmark errors

frouioui:
## Description

Small fix to change the way we calculate the `Improved by %` for the `Errors` field. If the recent version has `Errors` higher than the old version then the `Improved by %` column should be negative.

**Before**
image

**After**
image


GuptaManan100 (approved):
LGTM 💣

PR #267, opened 2021-08-18
Fix order on which we display microbenchmark results

frouioui:
## Description

This pull request fixes the rendering of microbenchmark results by reordering the column properly.


GuptaManan100:
What! we were displaying results in the incorrect order?

frouioui:
> What! we were displaying results in the incorrect order?

Just found out, yes. Fixed the way we compute the difference and how we render the different values. Mostly, the two cols (ref and cmp) got mixed up

GuptaManan100:
You might need to update the tests too, reflect the changes there

GuptaManan100 (approved):
LGTM 🔎

PR #269, opened 2021-08-25
Add golang version to the execution

frouioui:
## Description

This pull request adds a field in the execution to set the golang version we want to use. The golang version is saved in a new column of the `execution` table and is then sent to Ansible that will install the proper version on the benchmark's host.

This will allow running benchmarks using different golang versions without having to manually change the Ansible configuration.

Corresponding pull request on PSDB to apply schema changes: https://app.planetscale.com/vitess/arewefastyet/deploy-requests/3


GuptaManan100 (approved):
🚢  it

PR #270, opened 2021-08-25
Not rendering results if their benchmark is not marked as finished

frouioui:
## Description

This pull request avoids the rendering of results if their benchmark is not marked as `finished`.


GuptaManan100 (approved):
Lgtm

PR #283, opened 2021-11-22
Split the Ansible playbook in several reusable playbooks

frouioui:
## Description

This pull request changes the main `macrobench` and `microbench` playbooks in multiple reusable pieces. Which includes: a playbook to setup a Vitess cluster, a playbook to do the initial setup of a benchmark node, a playbook to teardown a Vitess cluster, a playbook to clean the benchmark node after a macrobenchmark. These new playbook are made to be used over and over on the same benchmark node. Running the macrobench/microbench plays will teardown the previous Vitess cluster (if it is still around) and start from a clean slate.


ansible/macrobench.yml
```diff
@@ -59,44 +36,49 @@
     gateway_targets: "{% set targets = [] -%}{% for host in groups['vtgate'] -%}{% for gateway in hostvars[host]['gateways'] -%}{{ targets.append( hostvars[host]['ansible_default_ipv4']['address']+':'+ (gateway.port | default(15001) | string )) }}{% endfor -%}{% endfor -%}{{ targets }}"
     vtctld_targets: "{% set targets = [] -%}{% for host in groups['vtctld'] -%}{{ targets.append( hostvars[host]['ansible_default_ipv4']['address']+':'+ (vtctld_port | default(15000) | string )) }}{% endfor -%}{{ targets }}"
     node_targets: "{% set targets= [] -%}{% for host in groups['all'] -%}{{ targets.append( hostvars[host]['ansible_default_ipv4']['address']+':9100' ) }}{% endfor -%}{{ targets }}"
-    prometheus_version: 2.26.0
```
GuptaManan100, line 62: do we no longer need `prometheus_version`. Without this it would default to the latest right?

ansible/README.md
```diff
@@ -1,76 +0,0 @@
-## Ansible Provisioner for Vitess
-
-### Testing individual roles
-
-Most roles in the playbook have been configured to be tested with molecule. These can be run by issuing
```
GuptaManan100, line 5: Do we want to keep the readme and update it instead of removing it entirely?

GuptaManan100 (approved):

ansible/macrobench.yml
frouioui: it defaults to `2.27.0`  (https://github.com/cloudalchemy/ansible-prometheus#role-variables)

ansible/README.md
frouioui: thank you, it should not be removed entirely, I am still due for a small update on the old files/tests

PR #291, opened 2022-05-05
Decrease overall result variation in benchmarking results

frouioui:
This Pull Request contains various enhancements and bug fixes that improve the overall stability of the benchmarking results.

The list of things that are modified is:
- Temporarily removed the molecule tests from the CI. Better end-to-end tests will be written in a follow-up PR.
- Changed the macro-benchmark topology from `6 vtgates - 2 vttablets` to `3 vtgates - 4 vttablets (primary/replica in each shard)`
- Removed debugging in Ansible tasks
- Make sure the `vitess` user and group are created in the new benchmarking server
- Use the same storage device during micro and macro benchmarks
- Format properly the storage device between each benchmark
- Make sure to kill the `arewefastyetcli` processes before and after a benchmark
- Make sure to kill all `sysbench` processes before and after a benchmark
- Fixed equality issue with `ansible_os_family`
- Added support for `go1.18.1`
- Support the new flag syntax of `vtctlclient`
- Removed the `-gateway_implementation` flag from vtgate's service configuration
- Changed the default OLTP and TPCC sysbench configuration to have smaller and more reliable benchmarks
- Addition of a new CRON schedule for the tags/releases benchmarks
- Increased the timeout for microbenchmarks from 2 to 4 hours
- Fixed an issue in the results compare where only one result was compared with the current benchmark instead of all the results that need to be compared
- Avoid the duplication of Pull Request CRON benchmarks
- Fixed the main `cron` web page to show results from `today` (included) to `today - 31 days`
- Capped the minimum version of Vitess for tags/releases benchmark to `v12.0.0`
- Avoid running profiling for micro-benchmarks unless told otherwise by a newly introduced flag
- Fixed configuration issue in influxdb's ansible role
- Encrypt Grafana's secrets in Ansible


go/tools/git/git.go
```diff
@@ -135,7 +135,7 @@ func GetLatestVitessReleaseCommitHash(repoDir string) ([]*Release, error) {
 	}
 	var latestReleases []*Release
 	for _, release := range allReleases {
-		if release.Number[0] > 6 {
+		if release.Number[0] >= 12 {
```
GuptaManan100, line 138: We should at some point automate this to last 3 releases or something like that, instead of a hard-coded number. But this is fine for now too.

GuptaManan100 (approved):
Nothing else jumps out to me. Good to merge 👍

harshit-gangal (approved):
The changelog looks really good. 
Approving based on it.

go/tools/git/git.go
harshit-gangal: I agree on that.

frouioui: agreed

PR #295, opened 2022-07-11
Remove the v3 planner from the list of execution and from the website

frouioui:
## Description

This Pull Request removes the execution of benchmarks using the `v3` planner, instead only `Gen4` benchmarks will be run (instead of `v3` & `Gen4Fallback`).

The visualization of results for both `v3` and `Gen4` is removed. Now we can only see the results for `Gen4`.

### DIFF
--- a/go/server/cron.go
+++ b/go/server/cron.go
@@ -93,9 +93,9 @@ func (s *Server) createCrons() error {
 
 func (s *Server) getConfigFiles() map[string]string {
 	configs := map[string]string{
-		// "micro": s.microbenchConfigPath,
-		// "oltp":  s.macrobenchConfigPathOLTP,
-		// "tpcc":  s.macrobenchConfigPathTPCC,
+		"micro": s.microbenchConfigPath,
+		"oltp":  s.macrobenchConfigPathOLTP,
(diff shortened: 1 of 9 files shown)


go/server/cron.go
```diff
@@ -93,9 +93,9 @@ func (s *Server) createCrons() error {
 
 func (s *Server) getConfigFiles() map[string]string {
 	configs := map[string]string{
-		"micro": s.microbenchConfigPath,
-		"oltp":  s.macrobenchConfigPathOLTP,
-		"tpcc":  s.macrobenchConfigPathTPCC,
+		// "micro": s.microbenchConfigPath,
+		// "oltp":  s.macrobenchConfigPathOLTP,
+		// "tpcc":  s.macrobenchConfigPathTPCC,
```
GuptaManan100, line 98: I think we don't want to commit this? 🤣

GuptaManan100 (approved):
Lint failures are still there. Please merge after fixing the below 👇

go/server/cron.go
frouioui: We dont :)

PR #298, opened 2022-07-22
Rework how the Vitess flags are handled

frouioui:
## Description

This Pull Request mainly updates how we set the Vitess flags in Ansible. To do so, a new variable is introduced in the execution (`VitessVersionName`) to track which underlying version of Vitess we are benchmarking. This variable will then help us set the Vitess flags in Ansible dynamically.

This is motivated by the fact that different versions of Vitess have different requirements regarding the flags and their syntax. We saw errors on the most recent versions of Vitess (`cron` and certain `cron_pr`+`cron_pr_base`) where we were using VTTablet flags that got removed. On older versions of Vitess, such as release-13.0, we saw errors with the `-planner-version` flag being incorrect.

On top of this main change, this Pull Request also removes the schema tracking flag, as the schema tracking feature is enabled by default in v14. Better management of Ansible's extra vars is also brought by this Pull Request, all the keys are centralized in a single place, and a new method is introduced to set the extra vars. The tokenization of the Ansible inventory is also removed and replaced by the use of extra vars.


go/exec/exec.go
```diff
@@ -300,45 +348,55 @@ func (e *Exec) handleStepEnd(err error) {
...
-	if err != nil {
-		return nil, err
+	// Release branches / Tags
+	if strings.HasPrefix(e.Source, SourceReleaseBranch) || strings.HasPrefix(e.Source, SourceTag) {
+		matches := matchRelease.FindStringSubmatch(e.Source)
+		if len(matches) < 2 {
+			return fmt.Errorf("unkown format to find the version of vitess: %v", matches)
+		}
+		e.VitessVersionName = VitessPreviousVersion + matches[1]
+		return nil
```
frouioui, line 397: remove this return

GuptaManan100 (approved):

PR #306, opened 2022-09-14
Add the OLTP-SET benchmark

frouioui:
This PR refactors how we deal with multiple benchmarks. It is now possible to easily add a new macro benchmark.

The OLTP-SET benchmark is also added.


harshit-gangal (approved):
merge once unit test and lint test pass.

PR #308, opened 2022-09-15
Customise the Vitess configuration for a benchmark

frouioui:
This PR changes the execution pipeline to calculate which version of Vitess a benchmark will run. This information allows us to customize the Vitess configuration per version. We can now have specific flags for vttablet and vtgate depending on which version of Vitess will be run. These flags are passed through the benchmark yaml file.

The `exec-vitess-config` flag achieves this. This flag is a map that contains a map: `map[string]map[string]string`.
The outer map represents the different releases of Vitess and the inner map the different
binaries. Finally, the key of the inner map corresponds to the flags to add.
The key for the inner map is the Vitess version it has to be formatted using `-` to
separate the major/minor/patch numbers. Like so: `v15.0.0` becomes: `15-0-0`.

We can have as many releases as we want in the outer map. Only the one that is the closest
to this execution's version will be used. For instance, if we define two versions, `v14.0.0`
and `v15.0.0`, but the execution runs on Vitess `v14.0.3`, then the configuration for `v14.0.0`
will be used.

Here is an example of how the flag should be formatted in YAML:
```yaml
exec-vitess-config:
  14: # will match >= v14.0.0
    vtgate: --my_flag=0
  15: # will match >= v15.0.0 and override the flags defined in < 15.0.0
    vtgate: --my_flag=1
    vttablet: --my_flag=1
  15-0-1: # will match >= v15.0.1 and override the flags defined in < 15.0.1
    vtgate: --my_flag=3
    vttablet: --my_flag=3 --custom-flag=on
```


harshit-gangal:
what you feel about having like this

```
vtgate-exec-vitess-config:
  15-0-1:  --my_flag=3
vttablet-exec-vitess-config:
  15-0-1: --my_flag=3 --custom-flag=on
```

go/exec/vitess_config.go
```diff
@@ -0,0 +1,124 @@
...
+	rawVitessVersionConfig struct {
+		version git.Version
+		value   map[string]interface{}
+	}
+)
+
+func (e *Exec) prepareVitessConfiguration() error {
+	if len(e.rawVitessConfig) == 0 {
+		return nil
+	}
```
harshit-gangal, line 45: adding a test for this file will be great to see if this works as expected.

harshit-gangal (commented):
nice enhancement. left few comments.

harshit-gangal (approved):

go/exec/vitess_config.go
frouioui: done

PR #311, opened 2022-09-23
Addition of a flag to exclude some source from the crons

frouioui:
This PR adds a new flag to the server to exclude certain sources from the list of benchmarks we execute in the cron. `"cron_tags_14.0.0-rc1"` is removed as it does not build due to incorrect flags.


harshit-gangal:
I think along the same line,
we should benchmark only major version and last minor version and ignore anything in between

harshit-gangal (approved):

PR #322, opened 2022-10-06
Removed deprecated vttablet dbconfig flags

frouioui:
This PR replaces the deprecated dbconfig flags by the new proper flags for latest and v15 benchmarks.

### DIFF
--- a/ansible/roles/vttablet/templates/vttablet@.service.j2
+++ b/ansible/roles/vttablet/templates/vttablet@.service.j2
@@ -22,17 +22,10 @@ LimitMEMLOCK=infinity
 WorkingDirectory={{vitess_root}}
 User={{vitess_user}}
 ExecStart=/bin/bash -c 'vttablet \
-	{% if vitess_version_name == "latest" or vitess_version_name == "v_15" %}
-		--db_app_user=vt_app \
-		--db_dba_user=vt_dba \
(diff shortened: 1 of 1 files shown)


ansible/roles/vttablet/templates/vttablet@.service.j2
```diff
@@ -22,12 +22,19 @@ LimitMEMLOCK=infinity
...
+		--db_app_user=vt_app \
+		--db_dba_user=vt_dba \
+		--db_filtered_user=vt_filtered \
+		--db_repl_user=vt_repl \
+	{% else %}
+		--db-config-app-uname=vt_app \
+		--db-config-dba-uname=vt_dba \
+		--db-config-filtered-uname=vt_filtered \
+		--db-config-repl-uname=vt_repl \
+	{% endif %}
```
GuptaManan100, line 35: You don't need the if, else. The parameters `db_app_user` are very old. You can just use them.

GuptaManan100 (approved):

ansible/roles/vttablet/templates/vttablet@.service.j2
frouioui: gotcha

frouioui: done via e8d448a

PR #344, opened 2023-05-11
upgrade to use `go1.20.4` for all benchmarks

frouioui:


.github/workflows/deploy.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v1
```
dbussink, line 15: Let's also make sure to bump all these to `actions/setup-go@v4` to fix deprecation warnings.

dbussink (approved):

PR #369, opened 2023-06-06
Status page desktop version 

Camillemtd:
I have created a page status that consists of two components: "Previous Executions" and "Execution Queue," which are displayed as two tables on the page. To achieve this, I utilized two libraries: moment.js for date formatting and react-spinner to display a loading indicator while waiting for the API response.

### DIFF
--- a/website/.env.dev
+++ b/website/.env.dev
@@ -1,2 +1,2 @@
-VITE_API_QUEUE_URL=http://localhost:9090/api/queue
-VITE_API_RECENT_URL=http://localhost:9090/api/recent
+VITE_API_URL=http://localhost:9090/api/
+
--- a/website/.env.dev
+++ b/website/.env.dev
@@ -1,0 +1,2 @@
+VITE_API_QUEUE_URL=http://localhost:9090/api/queue
+VITE_API_RECENT_URL=http://localhost:9090/api/recent
--- a/website/src/components/ExecutionQueue/ExeQueue.jsx
+++ b/website/src/components/ExecutionQueue/ExeQueue.jsx
@@ -1,3 +1,19 @@
+/*
+Copyright 2023 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 import React from 'react';
 
 import '../ExecutionQueue/exeQueue.css'
--- a/website/src/components/ExecutionQueue/ExeQueue.jsx
+++ b/website/src/components/ExecutionQueue/ExeQueue.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+
+import '../ExecutionQueue/exeQueue.css'
+const ExeQueue = ({data}) => {
+    return (
+        <tr>
+            <td><a target='_blank' href={`https://github.com/vitessio/vitess/commit/${data.git_ref}`}>{data.git_ref.slice(0, 8)}</a></td>
+            <td>{data.source}</td>
+            <td>{data.type_of.slice(0, 4)}</td>
+            <td>{data.pull_nb}</td>
+        </tr>
+    );
+};
+
+export default ExeQueue;
--- a/website/src/components/ExecutionQueue/ExeQueue.jsx
+++ b/website/src/components/ExecutionQueue/ExeQueue.jsx
@@ -16,7 +16,7 @@ limitations under the License.
 
 import React from 'react';
 
-import '../ExecutionQueue/exeQueue.css'
+
 const ExeQueue = ({data}) => {
     return (
         <tr>
--- a/website/src/components/ExecutionQueue/ExeQueue.jsx
+++ b/website/src/components/ExecutionQueue/ExeQueue.jsx
@@ -20,9 +20,9 @@ import React from 'react';
 const ExeQueue = ({data}) => {
     return (
         <tr>
-            <td><a target='_blank' href={`https://github.com/vitessio/vitess/commit/${data.git_ref}`}>{data.git_ref.slice(0, 8)}</a></td>
+            <td><a target='_blank' href={`https://github.com/vitessio/vitess/commit/${data.git_ref}`}>{data.git_ref.slice(0, 6)}</a></td>
             <td>{data.source}</td>
-            <td>{data.type_of.slice(0, 4)}</td>
+            <td>{data.type_of}</td>
             <td>{data.pull_nb}</td>
         </tr>
     );
--- a/website/src/components/PreviousExecutions/PreviousExe.jsx
+++ b/website/src/components/PreviousExecutions/PreviousExe.jsx
@@ -1,3 +1,19 @@
+/*
+Copyright 2023 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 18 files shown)


website/src/components/ExecutionQueue/ExeQueue.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+import React from 'react';
+
+
+const ExeQueue = ({data}) => {
+    return (
+        <tr>
+            <td><a target='_blank' href={`https://github.com/vitessio/vitess/commit/${data.git_ref}`}>{data.git_ref.slice(0, 8)}</a></td>
+            <td>{data.source}</td>
+            <td>{data.type_of.slice(0, 4)}</td>
```
frouioui, line 25: We should display the full `type_of` field. Some benchmark types are longer than four chars, for instance: `oltp-readonly-olap`.

website/src/components/PreviousExecutions/PreviousExe.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
frouioui, line 1: Missing license header here

website/src/components/PreviousExecutions/PreviousExe.jsx
```diff
@@ -0,0 +1,47 @@
...
+    const getStatusClass = (status) => {
+        if (status === 'finished') {
+          return 'finished';
+        } else if (status === 'failed') {
+          return 'failed';
+        } else if (status ==='started') {
+            return 'started';
+        } else {
+            return 'default';
+        }
```
frouioui, line 18: IMHO, in this situation it is better to use a `switch` statement rather than a `if/if else/else` statement. Using a `switch` will greatly improve readability.

frouioui:
```js
switch (status) {
  case 'finished':
    return 'finished';
  case 'failed':
    return 'failed';
  case 'started':
    return 'started';
  default:
    return 'default';
}

```

frouioui: The reason why using a `switch` is better than `if/if else/else` is because your conditions are "simple" (no `OR`s / `AND`s, and usage of `===`) and always on the same variable (`status`).

website/src/components/PreviousExecutions/PreviousExe.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+        //Create a Moment object from the given date string
+        
+        const startedDate = moment(data.started_at)
+        const finishedDate = moment(data.finished_at)
+        // Format the date using the format method of Moment.js
+        // Here, we use 'DD/MM/YYYY HH:mm:ss' as the desired format
+
+        const formattedStartedDate = startedDate.format('MM/DD/YYYY HH:mm')
+        const formattedFinishedDate = finishedDate.format('MM/DD/YYYY HH:mm')
```
frouioui, line 29:
```suggestion
        //Create a Moment object from the given date string
        const startedDate = moment(data.started_at)
        const finishedDate = moment(data.finished_at)

        // Format the date using the format method of Moment.js
        // Here, we use 'DD/MM/YYYY HH:mm:ss' as the desired format
        const formattedStartedDate = startedDate.format('MM/DD/YYYY HH:mm')
        const formattedFinishedDate = finishedDate.format('MM/DD/YYYY HH:mm')
```

website/src/components/PreviousExecutions/PreviousExe.jsx
```diff
@@ -0,0 +1,47 @@
...
+    return (
+
+        <tr>
+            <td>{data.uuid.slice(0, 8)}</td>
+            <td><a target='_blank' href={`https://github.com/vitessio/vitess/commit/${data.git_ref}`}>{data.git_ref.slice(0,6)}</a></td>
+            <td>{data.source}</td>
+            <td>{formattedStartedDate}</td>
+            <td>{formattedFinishedDate}</td>
+            <td>{data.type_of.slice(0,4)}</td>
```
frouioui, line 38: Same thing here, we should not trim the name of the benchmark's type. We should display the entire string.

website/src/components/PreviousExecutions/previousexe.css
```diff
@@ -0,0 +1,37 @@
+td{
+    padding: 20px 0px;
```
frouioui, line 2: License header is missing here too.

website/src/components/footer/footer.css
```diff
@@ -15,11 +15,14 @@ limitations under the License.
...
 
 .footer{
+    position: relative;
+    bottom: 0;
     width: 100%;
     height: 160px;
     background-color: #343A40;
     margin-top: 100px;
     align-items: center;
+
```
frouioui, line 25: nit: we can remove the new line

website/src/index.css
```diff
@@ -21,20 +21,38 @@ limitations under the License.
...
+
+.line{
+  border: 1px solid #B5ADAD;
+  margin-top: 100px;
+  margin-bottom: 100px;
+}
+
+.line2{
+  border: 1px solid #B5ADAD;
+  margin-top: 10px;
```
frouioui, line 57: `line2` is not a very self-informative name, especially when there is a `line` class right above. You should perhaps add a comment to explain how  `line` and `line2` are different

website/src/pages/Status/Status.jsx
```diff
@@ -15,12 +15,133 @@ limitations under the License.
 */
 
 import React from 'react';
+import { useState, useEffect } from 'react';
+
+import './status.css'
+
+import PreviousExe from '../../components/PreviousExecutions/PreviousExe';
+import ExeQueue from '../../components/ExecutionQueue/ExeQueue';
+import RingLoader from "react-spinners/RingLoader";
```
frouioui, line 24: I feel like this import should be move above, among the other two external imports.

website/src/pages/Status/Status.jsx
```diff
@@ -15,12 +15,133 @@ limitations under the License.
...
+        const responsePreviousExe = await fetch('http://localhost:9090/api/recent');
+  
+        const jsonDataQueue = await responseQueue.json();
+        const jsonDataPreviousExe = await responsePreviousExe.json();
+        
+        setDataQueue(jsonDataQueue);
+        setDataPreviousExe(jsonDataPreviousExe);
+        setIsLoading(false)
+      } catch (error) {
+        console.log('Erreur lors de la récupération des données de l\'API', error);
```
frouioui, line 46: Error should be in english. Moreover, we should ideally have some kind of fallback if the API fails. Instead of leaving the loader loading forever, we should show some text on the page informing the user that the load failed and inviting the user to try again.

website/src/pages/Status/Status.jsx
```diff
@@ -15,12 +15,133 @@ limitations under the License.
...
+  const [isLoading, setIsLoading] = useState(true)
+  const [dataQueue, setDataQueue] = useState([]);
+  const [dataPreviousExe, setDataPreviousExe] = useState([]);
+  
+  
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const responseQueue = await fetch('http://localhost:9090/api/queue');
+        const responsePreviousExe = await fetch('http://localhost:9090/api/recent');
```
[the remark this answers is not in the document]
frouioui: We have `dev` and `prod` environment, this code will not work in the `prod` environment. Using a configuration file or environment variable is a suggestion, but there are probably other alternative. At the end of the day, we should be able to distinguish if we are in a `dev` or `prod` environment and have different API hostname and port between the two environments.

frouioui (changes requested):
First pass of the PR done. Left several items that must be addressed before merging the PR.

website/.env.dev
```diff
@@ -0,0 +1,2 @@
+VITE_API_QUEUE_URL=http://localhost:9090/api/queue
+VITE_API_RECENT_URL=http://localhost:9090/api/recent
```
frouioui, line 2: I think it would be better to have only the domain + the port, i.e.:

```env
VITE_API=http://localhost:9090/api/
```

And then you can append the route in your code. That way we don't explicitly list all routes in a single file.

website/vite.config.js
```diff
@@ -1,7 +1,31 @@
...
+import dotenv from 'dotenv';
+
+
+export default defineConfig(({ mode }) => {
+  // Charge les variables d'environnement à partir du fichier .env correspondant au mode (dev ou prod)
+  dotenv.config({ path: `./.env.${mode}` });
+
+  return {
+    // Configuration de votre application Vite
+    // ...
```
frouioui, line 29: Comments should be in english!

frouioui (approved):

frouioui:
LGTM! Thanks @Camillemtd great work 🎉

PR #370, opened 2023-06-07
Home/Header responsive 

Camillemtd:
Media queries Tablets and mobile for Home Page and Header component.


website/src/components/header/Header.jsx
```diff
@@ -23,32 +23,97 @@ import './header.css'
 import logo from '../../assets/logo.png'
 
 const Header = () => {
+
+    const [openMenu, setOpenMenu] = useState(false)
+    const handleMenuClick = () => {
+        window.scrollTo(0, 0);
```
frouioui, line 29: If we are halfway through the web page, reading the page, and then someone clicks on the menu button, will it scroll the page to the very beginning?

Camillemtd: This is used to scroll the page to the top when clicking on a link in the menu.

frouioui (approved):

PR #375, opened 2023-06-12
[Feature] Added 404 Error Page for Non-Existent Routes

Camillemtd:
I have created an error page for when a person searches a non-existent route. They will encounter an error message with a 404 code and a link that redirects them to the Home page.

### DIFF
--- a/website/src/utils_/Error.jsx
+++ b/website/src/utils_/Error.jsx
@@ -24,7 +24,7 @@ import logo from '../assets/homeLogoLarge.png'
 const Error = () => {
     return (
         <div className='error'>
-           <h2>ERROR 404</h2>
+           <h2>404</h2>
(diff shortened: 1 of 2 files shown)


website/src/utils_/Error.jsx
```diff
@@ -15,11 +15,18 @@ limitations under the License.
...
+
+import '../utils_/error.css'
+import logo from '../assets/homeLogoLarge.png'
 
 const Error = () => {
     return (
-        <div>
-            error
+        <div className='error'>
+           <h2>ERROR 404</h2>
```
frouioui, line 27: I think just `404` here feels simpler

frouioui (approved):

frouioui (approved):

PR #393, opened 2023-06-23
 Implement Search Page with Macrobenchmark and Microbenchmark Tables

Camillemtd:
I have created the search page by reusing the concept of the Macrobenchmark table and the Microbenchmark table, which are two separate components.


website/src/components/SearchComponents/SearchMicro/SearchMicro.jsx
```diff
@@ -0,0 +1,53 @@
+import React, {useState} from 'react';
```
frouioui, line 1: Missing header here 😀

website/src/components/SearchComponents/SearchMicro/searchMicro.css
```diff
@@ -0,0 +1,8 @@
+.search__microbench__bottom{
```
frouioui, line 1: Missing header here too

frouioui (approved):
When we land on the search page and there is no macrobenchmark result, we get a `null` displayed in the middle of the page. It would be good to replace this by zero-values or else.
image


Otherwise looks good to me!!

PR #394, opened 2023-06-26
Implement Compare Page for Benchmark Comparison between Commits

Camillemtd:
Creating the "Compare" Page that Displays a Table of Macro Benchmarks and Microbenchmarks to Compare Two Commits.

Closes https://github.com/vitessio/arewefastyet/issues/360


website/src/pages/Compare/Compare.jsx
```diff
@@ -0,0 +1,272 @@
...
+                    <span>
+                            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer a augue mi.
+                            Etiam sed imperdiet ligula, vel elementum velit.
+                            Phasellus sodales felis eu condimentum convallis.
+                            Suspendisse sodales malesuada iaculis. Mauris molestie placerat ex non malesuada.
+                            Curabitur eget sagittis eros. Aliquam aliquam sem non tincidunt volutpat. 
+                    </span>
+                    
+                </div>
+                <figure className='compareStats'></figure>
```
frouioui, line 121: I think we can get rid of that small header section on both search and comparison.

website/src/components/Macrobench/Macrobench.jsx
```diff
@@ -26,8 +26,8 @@ const Macrobench = ({data, gitRefLeft, gitRefRight}) => {
             <table>
                 <thead>
                     <tr>
-                        <th><h4>{gitRefLeft}</h4></th>
-                        <th><h4>{gitRefRight}</h4></th>
+                        <th><h4>{gitRefLeft.slice(0, 20)}</h4></th>
+                        <th><h4>{gitRefRight.slice(0, 20)}</h4></th>
```
frouioui, line 30: Mhm... I don't think it is fair to trim the git ref here. I think the caller of the `Macrobench` component should be the one responsible for trimming it.

website/src/components/MacrobenchMobile/MacrobenchMobile.jsx
```diff
@@ -87,7 +87,7 @@ const MacrobenchMobile = ({data, gitRefLeft, gitRefRight, setCurrentSlideIndexMo
                     </SwiperSlide>
                     <SwiperSlide>
                         <div className='macrobench__data flex--column' >
-                            <h4>{gitRefRight}</h4>
+                            <h4>{gitRefRight.slice(0, 19)}</h4>
```
frouioui, line 90: Same applies here.

website/src/components/MacrobenchMobile/MacrobenchMobile.jsx
```diff
@@ -65,7 +65,7 @@ const MacrobenchMobile = ({data, gitRefLeft, gitRefRight, setCurrentSlideIndexMo
                     </SwiperSlide>
                     <SwiperSlide>
                         <div className='macrobench__data flex--column'>
-                            <h4>{gitRefLeft}</h4>
+                            <h4>{gitRefLeft.slice(0, 19)}</h4>
```
frouioui, line 68: Same applies here.

website/src/pages/Compare/Compare.jsx
```diff
@@ -0,0 +1,272 @@
...
+            <div className='compare__bottom'>
+                <div className='justify--content form__container'>
+                    <h3>Comparing <a href={`https://github.com/vitessio/vitess/commit/${gitRefLeft}`}>{gitRefLeft.slice(0, 8)}</a> with <a href={`https://github.com/vitessio/vitess/commit/${gitRefRight}`}>{gitRefRight.slice(0, 8)}</a></h3>
+                    <form className='justify--content'>
+                        <input
+                        type="text"
+                        value={gitRefLeft}
+                        onChange={handleInputChangeLeft}
+                        placeholder="Left commit SHA"
+                        className='form__inputLeft'></input>
```
frouioui, line 133: For both inputs we should have only a place holder, the field should be empty. It currently has `Left` and `Right` that we need to remove before being able to type code.

website/src/pages/Compare/Compare.jsx
```diff
@@ -0,0 +1,272 @@
...
+                        <input
+                        type="text"
+                        value={gitRefRight}
+                        onChange={handleInputChangeRight}
+                        placeholder="Left commit SHA"
```
frouioui, line 138: The placeholder here should be about the right commit SHA.

website/src/pages/Compare/Compare.jsx
```diff
@@ -0,0 +1,272 @@
...
+                            Curabitur eget sagittis eros. Aliquam aliquam sem non tincidunt volutpat. 
+                    </span>
+                    
+                </div>
+                <figure className='compareStats'></figure>
+            </div>
+            <figure className='line'></figure>
+            <div className='compare__bottom'>
+                <div className='justify--content form__container'>
+                    <h3>Comparing <a href={`https://github.com/vitessio/vitess/commit/${gitRefLeft}`}>{gitRefLeft.slice(0, 8)}</a> with <a href={`https://github.com/vitessio/vitess/commit/${gitRefRight}`}>{gitRefRight.slice(0, 8)}</a></h3>
```
frouioui, line 126: If both strings (left and right SHAs) are empty, then this will render: `Comparing with`

frouioui (changes requested):
Few comments to fix, otherwise looks nice! 🚀  One more step towards parity.

frouioui (approved):

Issue #390, opened 2023-06-22
Rework the entire UI using React and reach parity with the old UI


frouioui:
## Description

As explained in https://github.com/vitessio/arewefastyet/issues/328 we want to get rid of the old UI and create a new UI using React. The goal of this issue is to reach parity with the old UI.

## Frontend

Parity will be reached when we have the following pages implemented:

- [x] https://github.com/vitessio/arewefastyet/issues/359
  - https://github.com/vitessio/arewefastyet/pull/365
  - https://github.com/vitessio/arewefastyet/pull/367
  - https://github.com/vitessio/arewefastyet/pull/370
  - https://github.com/vitessio/arewefastyet/pull/375
- [x] https://github.com/vitessio/arewefastyet/issues/358
  - https://github.com/vitessio/arewefastyet/pull/369
  - https://github.com/vitessio/arewefastyet/pull/371
  - https://github.com/vitessio/arewefastyet/pull/372
  - https://github.com/vitessio/arewefastyet/pull/374
- [x] https://github.com/vitessio/arewefastyet/issues/352
  - https://github.com/vitessio/arewefastyet/pull/385
- [x] https://github.com/vitessio/arewefastyet/issues/355
  - https://github.com/vitessio/arewefastyet/pull/387
- [x] https://github.com/vitessio/arewefastyet/issues/360
- [ ] https://github.com/vitessio/arewefastyet/issues/362
- [ ] https://github.com/vitessio/arewefastyet/issues/363
- [x] https://github.com/vitessio/arewefastyet/issues/361

## Backend

Modifying the UI to use React means we need to develop a new API to fetch data from the new frontend to the database. You can find the tracking Pull Requests in the list below:

- https://github.com/vitessio/arewefastyet/pull/386
- https://github.com/vitessio/arewefastyet/pull/373
- https://github.com/vitessio/arewefastyet/pull/364
- https://github.com/vitessio/arewefastyet/pull/376
- https://github.com/vitessio/arewefastyet/pull/377
- https://github.com/vitessio/arewefastyet/pull/378
- https://github.com/vitessio/arewefastyet/pull/380

PR #396, opened 2023-06-28
Creation of the macrobenchQueryPlan page and its queryPlan component

Camillemtd:
A link is added to the macro page to navigate to a new page called 'macrobench query plan'. Then, I called the API and created a component for each SQL query to optimize the readability of the information.

Closes https://github.com/vitessio/arewefastyet/issues/390

### DIFF
--- a/website/src/components/QueryPlan/QueryPlan.jsx
+++ b/website/src/components/QueryPlan/QueryPlan.jsx
@@ -1,0 +1,92 @@
+/*
+Copyright 2023 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, {useState}from 'react';
+
+import '../QueryPlan/queryPlan.css'
+
+const QueryPlan = ({data}) => {
+
+    const [showPlan, setShowPlan] = useState(false);
+
+    const togglePlan = () => {
+        setShowPlan(!showPlan);
+    };
+    
+
+    return (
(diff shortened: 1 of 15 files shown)


website/src/components/QueryPlan/QueryPlan.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+const QueryPlan = ({data, isOpen, togglePlan}) => {
+
+    const queryPlanKeyStyle = {
+        background: isOpen && window.innerWidth > 1225 ? 'orange' : 'initial',
+      };
+    
+      const queryTitleStyle = {
+        background: isOpen && window.innerWidth < 1225 ? 'orange' : 'initial',
+      };
```
frouioui, line 31: We should use constant for hardcoded numbers like these or add a comment describing it - otherwise we do not know what it is

frouioui: There is an indentation issue with the second const here

website/src/pages/MacroQueriesCompare/MacroQueriesCompare.jsx
```diff
@@ -0,0 +1,77 @@
...
+*/
+
+import React, { useEffect, useState } from 'react';
+import { useLocation } from 'react-router-dom';
+
+import '../MacroQueriesCompare/macroQueriesCompare.css'
+
+import { errorApi } from '../../utils/Utils';
+import QueryPlan from '../../components/QueryPlan/QueryPlan';
+const MacroQueriesCompare = () => {
```
frouioui, line 24: A new line between the two here would be great

website/src/pages/MacroQueriesCompare/MacroQueriesCompare.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const { state } = useLocation()
+    const commitA = state.another
+    const commitB = state.more
+    const type = state.some.type
```
frouioui, line 29: `another`, `more`, `some` seems off here

frouioui (commented):
Outside of the few comments i have left, this is good!

frouioui (approved):
LGTM, let's just address https://github.com/vitessio/arewefastyet/pull/396/files#r1246953788 in a follow-up PR

PR #398, opened 2023-07-07
Introducing the useApiCall hook for frontend API calls

Camillemtd:
I have created a hook for API calls in the frontend. Here is a detailed explanation of how this hook works:

Dependency import: The hook utilizes the necessary dependencies from React, namely useState and useEffect, as well as importing the errorApi utility from the Utils file.

Hook definition: The hook is a function named useApiCall that takes a URL as a parameter. This URL is used to make the API call.

State initialization: The hook uses three states with useState.

data is initialized with an empty array ([]) and will be used to store the data retrieved from the API.
isLoading is initialized as true and indicates whether the data is currently being loaded from the API.
error is initialized as null and will be used to store any errors that occur during the data retrieval from the API.
useEffect usage: The hook uses useEffect to perform the API call when the URL changes or when the component is mounted. The callback provided to useEffect is an async function named fetchData.

The fetchData function performs the API call using the provided URL. It uses await fetch(url) to send a GET request to the URL and retrieves the response. Then, it uses await response.json() to extract the JSON data from the response.

State update: Once the data is retrieved, the fetchData function uses setData to update the data state with the obtained JSON data. It then uses setIsLoading to mark the loading as completed by setting isLoading to false.

Error handling: If an error occurs during the data retrieval, the fetchData function uses console.log to display the error message and uses setError to update the error state with the errorApi value. It also uses setIsLoading to mark the loading as completed by setting isLoading to false.

Initial fetchData call: When the hook is mounted or when the URL changes, the useEffect callback is executed, triggering the initial call to fetchData.

Return of values: At the end of the hook, an object containing the data, isLoading, and error values is returned. These values can be used in the component that uses the hook to access the retrieved data, loading state, and any potential errors.

The useApiCall hook encapsulates all the logic for API calls and state management, making it easy to reuse this functionality in different components of your application.

### DIFF
--- a/website/src/utils/Hook.jsx
+++ b/website/src/utils/Hook.jsx
@@ -18,6 +18,8 @@ import { useState, useEffect } from 'react';
 
 import { errorApi } from '../utils/Utils';
 
+//
+
 const useApiCall = (url) => {
   const [data, setData] = useState([]);
   const [isLoading, setIsLoading] = useState(true);
--- a/website/src/utils/Hook.jsx
+++ b/website/src/utils/Hook.jsx
@@ -1,0 +1,47 @@
+/*
+Copyright 2023 The Vitess Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


website/src/utils/Hook.jsx
```diff
@@ -0,0 +1,49 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { useState, useEffect } from 'react';
+
+import { errorApi } from '../utils/Utils';
+
+//
+
```
frouioui, line 22: nit: we can remove this

frouioui (approved):
LGTM, comment can be addressed in a follow-up PR

PR #403, opened 2023-07-18
Add Benchmark Statistics API Endpoint and Retrieval Functionality

Camillemtd:
This pull request introduces a new feature to retrieve benchmark statistics in the server's API. It includes updates to the codebase to fetch and provide benchmark statistics to the clients.

Changes Made:

Added a new struct benchmarkStats with three fields: Benchmark_total, Benchmark_finished, and Benchmark_30Days. This struct will hold the benchmark statistics.

Implemented the GetBenchmarkStats function which retrieves the benchmark statistics from the database using the provided SQLClient. It executes a SQL query to count the total number of benchmarks, the number of finished benchmarks, and the number of benchmarks within the last 30 days. The query results are then assigned to the respective fields of the benchmarkStats struct.

In the getStatusStats function of the Server struct, the GetBenchmarkStats function is called to fetch the benchmark statistics using the server's dbClient. If an error occurs during the retrieval process, an internal server error response is sent back to the client along with the error message. Otherwise, the benchmark statistics are returned as a JSON response with the status code 200 OK.

Reason for the Change:

The addition of benchmark statistics retrieval enhances the functionality of the server's API by providing valuable information about benchmarks. This allows clients to monitor and analyze the benchmarking process more effectively.

### DIFF
--- a/go/exec/status.go
+++ b/go/exec/status.go
@@ -29,34 +29,32 @@ const (
 	StatusFinished = "finished"
 )
 
-type benchmarkStats struct {
-	Benchmark_total    int
-	Benchmark_finished int
-	Benchmark_30Days   int
+type BenchmarkStats struct {
+	Total    int
+	Finished int
+	Last30Days   int
 }
 
-func GetBenchmarkStats(client storage.SQLClient) (benchmarkStats, error) {
-	rows, err := client.Select(`
-        SELECT
+func GetBenchmarkStats(client storage.SQLClient) (BenchmarkStats, error) {
+	rows, err := client.Select(`SELECT
             (SELECT COUNT(uuid) FROM execution) AS count_status,
             (SELECT COUNT(*) FROM execution WHERE status = 'finished') AS count_finished,
             (SELECT COUNT(*) FROM execution WHERE started_at >= DATE_SUB(CURDATE(), INTERVAL 30 DAY)) AS count_all
         FROM
             execution
-		LIMIT 1;
-    `)
+		LIMIT 1;`)
 
 	if err != nil {
-		return benchmarkStats{}, err
+		return BenchmarkStats{}, err
 	}
 
 	defer rows.Close()
 
-	var res benchmarkStats
+	var res BenchmarkStats
 	if rows.Next() {
-		err := rows.Scan(&res.Benchmark_total, &res.Benchmark_finished, &res.Benchmark_30Days)
+		err := rows.Scan(&res.Total, &res.Finished, &res.Last30Days)
 		if err != nil {
-			return benchmarkStats{}, err
+			return BenchmarkStats{}, err
 		}
 	}
 	return res, nil
--- a/go/exec/status.go
+++ b/go/exec/status.go
@@ -30,18 +30,18 @@ const (
 )
 
 type BenchmarkStats struct {
-	Total    int
-	Finished int
-	Last30Days   int
+	Total      int
+	Finished   int
+	Last30Days int
 }
 
 func GetBenchmarkStats(client storage.SQLClient) (BenchmarkStats, error) {
-	rows, err := client.Select(`SELECT
-            (SELECT COUNT(uuid) FROM execution) AS count_status,
-            (SELECT COUNT(*) FROM execution WHERE status = 'finished') AS count_finished,
-            (SELECT COUNT(*) FROM execution WHERE started_at >= DATE_SUB(CURDATE(), INTERVAL 30 DAY)) AS count_all
-        FROM
-            execution
+	rows, err := client.Select(`SELECT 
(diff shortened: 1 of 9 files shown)


go/exec/status.go
```diff
@@ -18,9 +18,46 @@
...
 	StatusStarted  = "started"
 	StatusFailed   = "failed"
 	StatusFinished = "finished"
 )
+
+type benchmarkStats struct {
+	Benchmark_total    int
+	Benchmark_finished int
+	Benchmark_30Days   int
+}
```
frouioui, line 36: The structure `benchmarkStats` should be exported as it is used as a return parameter of `GetBenchmarkStats` which is itself an exported function.

Moreover, the fields should follow the camelcase naming convention without underscores.

go/exec/status.go
```diff
@@ -18,9 +18,46 @@
...
+
+func GetBenchmarkStats(client storage.SQLClient) (benchmarkStats, error) {
```
frouioui, line 38: There is a mix of tabs and spaces indentation in this function, let's stick to tabs instead of mixing both

go/exec/status.go
```diff
@@ -18,9 +18,46 @@
...
+	rows, err := client.Select(`
+        SELECT
+            (SELECT COUNT(uuid) FROM execution) AS count_status,
+            (SELECT COUNT(*) FROM execution WHERE status = 'finished') AS count_finished,
+            (SELECT COUNT(*) FROM execution WHERE started_at >= DATE_SUB(CURDATE(), INTERVAL 30 DAY)) AS count_all
+        FROM
+            execution
+		LIMIT 1;
```
frouioui, line 46: I don't think the `LIMIT 1` is required here as `COUNT`s always return one row anyway

go/exec/status.go
```diff
@@ -18,9 +18,46 @@
...
+	defer rows.Close()
+
+	var res benchmarkStats
+	if rows.Next() {
+		err := rows.Scan(&res.Benchmark_total, &res.Benchmark_finished, &res.Benchmark_30Days)
+		if err != nil {
+			return benchmarkStats{}, err
+		}
+	}
+	return res, nil
```
frouioui, line 62: this is good

website/src/pages/Status/Status.jsx
```diff
@@ -36,6 +37,12 @@ const Status = () => {
     isLoading: isLoadingPreviousExe,
     error: errorPreviousExe,
   } = useApiCall(`${import.meta.env.VITE_API_URL}recent`);
+  const {
+    data: dataStatusStats,
+    isLoading: isLoadingStatusStats,
+    error: errorStatusStats,
+  } = useApiCall(`${import.meta.env.VITE_API_URL}status/stats`);
+  console.log(dataStatusStats);
```
frouioui, line 45: I think we can get rid of this `console.log` i am assuming it was added for debugging purposes

website/src/pages/Status/Status.jsx
```diff
@@ -52,7 +59,29 @@ const Status = () => {
...
 
-        <figure className="statusStats"></figure>
+        <div className="statusStats flex--column">
+          <CountUp
+            start={0}
+            end={dataStatusStats.Benchmark_total}
+            style={{ fontSize: "5rem", color: "#E77002" }}
+            duration={3}
+          />
+          <span className="countUp__title">Benchmark Total</span>
```
frouioui, line 69: `Benchmark Total` -> `Total Benchmarks`

frouioui, line 69: It would be nice to have the top of that div at the same level as the top of the `h2`

go/exec/status.go
```diff
@@ -29,34 +29,32 @@ const (
...
 
-type benchmarkStats struct {
-	Benchmark_total    int
-	Benchmark_finished int
-	Benchmark_30Days   int
+type BenchmarkStats struct {
+	Total    int
+	Finished int
+	Last30Days   int
 }
```
frouioui: `gofmt` is missing here

go/exec/status.go
```diff
@@ -29,34 +29,32 @@ const (
...
 
-func GetBenchmarkStats(client storage.SQLClient) (benchmarkStats, error) {
-	rows, err := client.Select(`
-        SELECT
+func GetBenchmarkStats(client storage.SQLClient) (BenchmarkStats, error) {
+	rows, err := client.Select(`SELECT
             (SELECT COUNT(uuid) FROM execution) AS count_status,
             (SELECT COUNT(*) FROM execution WHERE status = 'finished') AS count_finished,
```
frouioui: We need to change from spaces to tabs on line `40` to `44`

frouioui (approved):

PR #405, opened 2023-07-21
Improvements: Iterated on Each Page for Design and Code Enhancemen

Camillemtd:
I iterated on each page to improve the design and the code.


website/src/components/CRONComponents/Chart/Chart.jsx
frouioui, line 1: this is a nice optimization, it's important to reduce repetition in the codebase

website/src/components/CRONComponents/Chart/Chart.jsx
```diff
@@ -0,0 +1,109 @@
...
+limitations under the License.
+*/
+
+import React from "react";
+import { ResponsiveLine } from "@nivo/line";
+
+const ResponsiveChart = ({ data, title, colors, isFirstChart }) => {
+  return (
+    <div className="chart">
+      {data.length > 0 && (
```
frouioui, line 23: Should we return early from `ResponsiveChart` in case `data.length` is equal to zero? That would prevent us from creating an empty div

package-lock.json
frouioui, line 1: We can remove this file

website/src/components/SearchComponents/SearchMacro/SearchMacroDesktop/SearchMacroDesktop.jsx
```diff
@@ -0,0 +1,59 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React from "react";
+
+import "../SearchMacroDesktop/searchMacroDesktop.css";
+
+import { formatByteForGB, fixed } from "../../../../utils/Utils";
```
frouioui, line 21: nit: group imports together and/or leave a new line between the component definition and the last import

website/src/pages/Status/Status.jsx
```diff
@@ -146,7 +146,7 @@ const Status = () => {
                 <PreviousExe
                   data={previousExe}
                   key={index}
-                  className={backgroundGrey}
+                  // className={backgroundGrey}
```
frouioui, line 149: nit: this line can be removed

website/src/pages/CRON/CRON.jsx
```diff
@@ -214,6 +225,34 @@ const CRON = () => {
...
+    {
+      data: TPSData,
+      title: "TPS (Transactions per second)",
+      colors: ["#E77002"],
+    },
+    {
+      data: latencyData,
+      title: "Latency (Milliseconds)",
+      colors: ["#E77002"],
+    },
```
frouioui, line 243: nit: `#E77002` can be replaced by `Yellow`

frouioui (approved):
Left a few comments and nits for now. Good job at optimizing the CRON page, it used to be pretty lengthy but the code is looking neat now!

I'll review again once this is marked as ready for review

website/src/components/CRONComponents/Chart/Chart.jsx
```diff
@@ -18,91 +18,91 @@ import React from "react";
 import { ResponsiveLine } from "@nivo/line";
 
 const ResponsiveChart = ({ data, title, colors, isFirstChart }) => {
+  console.log(data)
```
frouioui, line 21: I think we can remove this console.log too

website/src/components/CRONComponents/CRONSummary/CronSummary.jsx
```diff
@@ -14,48 +14,51 @@ See the License for the specific language governing permissions and
...
+
+  const transformedData = [
+    {
+      id: "QPSTotal",
+      data: data.Data.map((item) => ({
+        x: item.CreatedAt,
+        y: item.QPSTotal,
+      })),
+    },
+  ];
```
frouioui, line 33: I am getting a `Cannot read properties of null (reading 'map')` error on line `28`, I think we should handle empty results in `CronSummary`

website/src/components/StatusComponents/PreviousExecutions/PreviousExe.jsx
```diff
@@ -14,30 +14,51 @@ See the License for the specific language governing permissions and
...
+          <span>{data.pull_nb}</span>
+        ) : (
+          <a
+            target="_blank"
+            href={`https://github.com/vitessio/vitess/pull/${data.pull_nb}`}
+          >
+            {data.pull_nb}
+          </a>
+        )}
+      </span>
```
frouioui, line 53: nit picking here, but instead of `0` let's have an empty span if there is no pull request attached to this benchmark.

website/src/components/CRONComponents/CRONSummary/CronSummary.jsx
```diff
@@ -14,48 +14,53 @@ See the License for the specific language governing permissions and
...
+
+  if (data == null) {
+    transformedData.push({
+      id: "QPSTotal",
+      data: data.Data.map((item) => ({
+        x: item.CreatedAt,
+        y: item.QPSTotal,
+      })),
+    });
+  }
```
frouioui, line 35: Mhm this seems odd to me. Shouldn't we check for data not null? Here, if data is null then we attempt to do a `.Data.map` on it

PR #424, opened 2023-08-08
Changed homepage logo and added drop-shadow to it

marsian83:
The homepage seems to have been using two different images for logo with shadow which was of bad quality. So I have changed it to use the original logo, and added a drop-shadow around it through css


frouioui (approved):

frouioui:
Thanks for your contribution @marsian83, the logo is looking great!

PR #425, opened 2023-08-08
Footer Section

kirtanchandak:
In this PR, I've developed the footer section!


website/src/components/footer/Footer.jsx
```diff
@@ -14,16 +14,82 @@ See the License for the specific language governing permissions and
...
-        <div className='footer justify--content'>
-            <span>Copyright © 2023</span>
+  const icons = [
+    { url: "https://github.com/vitessio/arewefastyet", icon: AiFillGithub },
+    { url: "https://vitess.io/slack", icon: AiFillSlackCircle },
+    { url: "https://twitter.com/vitessio", icon: AiFillTwitterCircle },
+    {
+      url: "https://stackoverflow.com/search?q=vitess",
+      icon: BsStackOverflow,
+    },
```
frouioui, line 36: Let's inline these fours lines too, just like we did in lines `30`, `31` and `32`

frouioui:
Moreover, it would be great if the footer was the same width as the header and the other components of the page. With this change it feels like the footer is inside an even smaller container.

kirtanchandak:
> Moreover, it would be great if the footer was the same width as the header and the other components of the page. With this change it feels like the footer is inside an even smaller container.

```
max-width: 100vw;
```
would also work, the current footer also has this

website/src/components/footer/Footer.jsx
kirtanchandak: okay

frouioui (approved):
LGTM @kirtanchandak thank you for this contribution!

PR #437, opened 2023-08-14
Error page added

HetviSoni:
## Description
I added the error page. Demo:

https://github.com/vitessio/arewefastyet/assets/76479104/b3805f46-a1d7-4e01-bde6-09b8217ef36d

## Responsive
The dark/light toggle button that I added in my last pr was not responsive. In small screen the position of the button would be in the middle of the screen. So I corrected its position.


website/src/components/header/Header.jsx
```diff
@@ -135,7 +128,15 @@ const Header = () => {
             <figure className="lineMenuBottom"></figure>
           </figure>
         )}
+        
       </div>
+      <button className="buttonMode" onClick={handleButtonClick}>
```
Camillemtd, line 133: I think it would be better if the mode switch button were to the left of the hamburger menu on mobile. Otherwise, I really like the button!

Camillemtd (commented):
The error page is very good, but the image seems to be of low quality, especially on a large screen where it becomes blurry.

website/src/components/header/Header.jsx
HetviSoni: I have made these changes.

HetviSoni:
Yup, I agree with you. I looked for some tools online to improve the quality but it's not working as expected. Can you suggest what should I do to improve the quality of this .png image?

marsian83:
You could use a tool like Adobe Photoshop to vectorize the Image, that way you can have an svg which won't lose its quality regardless of image size. Another thing you can do is try AI platforms which help upscale an Image. If that still doesn't help you might wanna consider using another Image altogether.

HetviSoni:
Hey @marsian83 I have messaged you for help on Slack. Do reply when you are free

HetviSoni:
Hey @Camillemtd the previous image that I used was of really bad quality. So I changed the image. This looks good at all screen sizes.
image
Shall I push this?

Camillemtd:
Hey @HetviSoni, yes, perfect, you can push this!

HetviSoni:
I have pushed the changes👍

Camillemtd (approved):
Perfect, the error page is much better this way! Thank you for your contribution!

PR #438, opened 2023-08-14
Added ui enhancement to the search bar in search

marsian83:
I have redesigned the search bar in the search page for the frontend.  I have also added a small focus animation on the search bar.


Camillemtd (approved):
I really like what you've done with the focus and the gradient of colors that nicely recalls the one on the homepage! So, we should do the same thing for the 'Compare' page soon. Thanks for your work!

marsian83:
Yes I thought of getting a approval for this and later change those as well, I will be doing it now

PR #441, opened 2023-08-19
Added Ui enhancement to the input bar on compare page

marsian83:
Added the same focus effect as of Search page input bar, on compare page input bars

### DIFF
--- a/website/src/pages/Compare/Compare.jsx
+++ b/website/src/pages/Compare/Compare.jsx
@@ -50,23 +50,21 @@ const Compare = () => {
     data: dataMacrobench,
     isLoading: isMacrobenchLoading,
     error: macrobenchError,
-    textLoading: macroTextLoading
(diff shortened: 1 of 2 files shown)


Camillemtd (commented):
I think it would be a good idea to keep the "input left" and "input right" classes for their specific CSS, but create a new "input" class that combines all the CSS properties they have in common. Currently, we already have three input elements with repetitive CSS. By doing this, if we need to implement new ones in future features, it will be much simpler. What do you think about this?

marsian83:
Yes you are correct, Actually just the border-radius was different in both of those element, so I have used the method of nth-child in css to achieve the required style

Camillemtd (approved):
Yes, perfect. I think in the future we'll need to do the same thing for the search bar where there are repetitions with the compare page. Thank you very much!

PR #443, opened 2023-08-28
Implementation of standardizing logs

Camillemtd:


frouioui:
Closing this for now as it is stale. We can reopen in the future if we want to.

PR #511, opened 2024-02-02
Revert "add GOMEMLIMIT to vttablet configuration" for v18

frouioui:
This introduced weird behavior in the benchmarks, reverting for now until i find a proper long term solution:
image


frouioui:
This works well for `>= v19` let's keep it. We should however revert this only for `v18` as this release needs more GOMEMLIMIT

Issue #515, opened 2024-03-04
Implement Statistical Analysis for all Benchmarks


frouioui:
## Description

To improve the reliability of our benchmarks and the trust we have in them, we should implement proper statistical analysis of our different benchmarking results. This method will allow us to aggregate the results of multiple benchmark runs for the same commit and config to provide more accurate results and comparisons.

To highlight the lack of reliability of our benchmarks here is a screenshot of 4 results (each row) for the exact same benchmark (same vitess commit, version of arewefastyet, workload, flags, etc). We can see a (non)impressive `~1.5%` variation between the first benchmark and the second benchmark in `total_qps`.
![image](https://github.com/vitessio/arewefastyet/assets/35779988/d5153eaf-d603-48e7-9ef7-79a3fde949b9)

## Current implementation

Currently arewefastyet benchmarks a commit only once per workload (TPC-C, OLTP, etc...). This run usually lasts for about 23 minutes including the setup and teardown of the host server. The result of this run is then stored in the database and used to do comparisons with other commits/tags.

When a benchmark starts we prepare the host machine, which takes about 6 minutes, we clean up everything (disk, processes, vitess cluster, data, etc), we install the binaries (vitess + arewefastyet), we build them, we setup the vitess cluster. Then benchmarking workload runs for approximately 12 minutes (preparation, warm up, benchmark). Finally, we clean up the host server by repeating some of the steps from the preparation phase : clean up of disk, processes, vitess cluster, etc, this step takes about 5 minutes to complete.

To compare two benchmarks we currently fetch all the results for a given configuration (commit + workload + reason of the benchmark) average them all, and do the same thing for the benchmark we want to compare against. Once we have our two averaged benchmark results, we calculate the % difference between the first and second results. And we get the results that we can see on: https://benchmark.vitess.io/macro?ltag=18.0.2&rtag=main.

## Desired implementation

In order to do proper statistical analysis we must run the same benchmark for the same commit and workload multiple times. Running it 10 times will provide us with a good enough p-value. However, running the same benchmark 10 times would take `~23 mins * 10 = 3h48m`, which is not even considerable given that on a normal day we can run more than 20 different benchmarks.

[...]

PR #517, opened 2024-03-14
Implementation of statistical analysis for all macro-benchmarks

frouioui:
## Problematic

This Pull Request is the first (big) step towards fixing the very long standing issue in `arewefastyet`: **reliability**. Since the project was created we ran lengthy benchmarks (25-35 mins) and ran them only once per config (SHA + workload + benchmark reason). Leaving us with potentially impaired results and no ability to distinguish the noise from the actual results. It was common for arewefastyet to observe a ±1.5% variation in results for the exact same benchmark, which in the long run is not okay. More information on the problem in https://github.com/vitessio/arewefastyet/issues/515.

## Proposed change

This Pull Request changes the entire benchmarking process: from when we spawn new benchmarks in the queue to when we render results on the UI. Along the way I made some minor refactoring and removed old and/or unused pieces of code such as: notification (was broken and will re-implement it later), some old tests, HTTP argument names, etc.

The _main_ change in this Pull Request is that we are now running the exact same benchmark `MaximumBenchmarkWithSameConfig` times (constant set to `10`) and we now use statistical analysis to observe and compare all of our results. 

### Queue

When the the CRON's handlers spawn new benchmarks they get added to the queue, before being added we multiply the benchmark by `MaximumBenchmarkWithSameConfig` minus the number of same benchmarks already being executed, in the queue and stored in the database as `finished`.

### Execution

After these new benchmarks are added to the queue, they get consumed by the execution component. Now we add two new labels to Ansible: `KeyLastIsSame` and `KeyNextIsSame`, that allow us to optimize how we prepare and clean up the host machine with Ansible. `KeyLastIsSame` is set if the previous benchmark is the same as the current one in which case we don't have to prepare the host, and `KeyNextIsSame` allows us to skip the clean up at the end of the benchmark if the next benchmark is the same. The time spent on sysbench also changed from 600 seconds to 60 seconds, allowing us to re-run a benchmark in less than 2-3 minutes.

Not much else change in the execution.

### Results

The API has changed quite a lot, we used to fetch all the results for a config and doing the median on all the results found. Instead, we now get all the results for a given benchmark and perform simple operations to get some statistics on the results: median, high, low, a 95% confidence interval. If we are just looking to get the results of one benchmark, we return those stats. Otherwise, if we want to compare two benchmarks, we take the two samples and their stats and perform statistical analysis using the Mann Whitney U test. From this, we get a delta in %, a P value and an Alpha value, enabling us to say if we can make statistical inference from the two samples. We are relying on some of the functions used by `benchmath` for this whole process.

### Website

The website has been changed to reflect all the new measurements and changes done to the API. Here is what the new comparison table looks like when comparing `v19.0.1` and a recent commit on Vitess' `main`:

image

## Related Issue

Resolves https://github.com/vitessio/arewefastyet/issues/515


ansible/macrobench.yml
```diff
@@ -35,7 +38,7 @@
   tags:
     - prometheus
   roles:
-    - { role: cloudalchemy.prometheus, when: stats_remote_db_host is defined }
+    - { role: cloudalchemy.prometheus, when: stats_remote_db_host is defined and arewefastyet_last_exec_is_same is undefined }
```
frouioui, line 41: There is a linter issue here on Ansible, which is fine to ignore. Been a long standing issue.

ansible/roles/macrobench/tasks/main.yml
```diff
@@ -13,6 +13,7 @@
 - name: Install arewefastyet
   become: yes
   become_user: root
+  when: arewefastyet_last_exec_is_same is undefined
```
notfelineit, line 16: This is my first time looking at this, could you explain what this does in layman's terms 😄

go/exec/exec.go
```diff
@@ -156,9 +167,17 @@ const (
 
 // NewExec creates a new *Exec with an autogenerated uuid.UUID as well
 // as a constructed infra.Infra.
```
notfelineit, line 169: I guess that means this comment can be updated with "autogenerated uuid.UUID OR provided UUID"? (If I understand correctly)

go/server/cron.go
```diff
@@ -57,6 +60,12 @@ var (
 	queue            executionQueue
 )
 
+func (ei executionIdentifier) equalWithoutUUID(id executionIdentifier) bool {
+	ei.UUID = ""
+	id.UUID = ""
```
notfelineit, line 65: Would this permanently set the UUIDs to empty string once `equalWithoutUUID` is called? Don't have the context on whether or not that is intentional!

website/components.json
```diff
@@ -6,7 +6,7 @@
   "tailwind": {
     "config": "tailwind.config.ts",
     "css": "assets/styles/tailwind.css",
-    "baseColor": "neutral",
+    "baseColor": "slate",
```
notfelineit, line 9: 🎨

website/src/common/Macrobench.tsx
```diff
@@ -42,197 +43,217 @@ export default function Macrobench({ data, gitRef, commits }) {
...
+  if (range.infinite == true) {
+    return "∞"
+  }
+  if (range.unknown == true) {
+    return "?"
+  }
+  return "±"+fixed(range.value, 1)+"%"
+}
+
+function Row({ title, oldVal, newVal, delta, insignificant, p, fmt }) {
```
notfelineit, line 218: This is nit, but since this is a tsx file anyway, could add types to this 😄

website/src/pages/ComparePage/ComparePage.tsx
```diff
@@ -80,65 +72,17 @@ export default function Compare() {
...
+                    commits={{ old: gitRef.old, new: gitRef.new }}
                   />
                 </div>
               );
             })}
           </div>
         </section>
       )}
-
-      {/*{!isMicrobenchLoading && dataMicrobench && (*/}
```
notfelineit, line 95: Could we delete this?

notfelineit (approved):
The math & golang look fine to me! I don't have greater context on arewefastyet though 😄

go/server/cron.go
notfelineit: Ha! You're totally right. I need to brush off more golang.

website/src/pages/ComparePage/ComparePage.tsx
frouioui: Yes, this whole comment was not used and not maintained. We disabled **micro** benchmarks a while ago. We should re-think them entirely if we want to re-implement them. This is just some clean up deletion.

ansible/roles/macrobench/tasks/main.yml
frouioui: Sure! On the snippet you highlighted we are checking if the last execution/benchmark _**is not**_ the same (same commit, same source/trigger and same workload). If the condition is true (last benchmark is different than the current one) we would execute the whole `Install arewefastyet` instruction.

The keyword `arewefastyet_last_exec_is_same` is set by the `exec` package when [preparing Ansible](https://github.com/vitessio/arewefastyet/blob/5053153cdaeb44537f44b4f2e32a65e43110eb99/go/exec/exec.go#L372) as an "[extra vars](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#defining-variables-at-runtime)". On the snippet below, if the last execution/benchmark **_is_** the same, then we set the "extra vars" `arewefastyet_last_exec_is_same` and then it will not longer be `undefined` when running Ansible and thus the whole `Install arewefastyet` step can be skipped.

```go
        if e.PreviousBenchmarkIsTheSame {
                e.AnsibleConfig.AddExtraVar(ansible.KeyLastIsSame, true)
        }
```

This whole thing is just to optimize how we run Ansible and avoid unnecessary steps, reducing the time needed to setup and cleanup the host machine from 10+ minutes to only 1-2 minutes.

frouioui: I hope that makes sense, let me know if it is unclear :)

website/src/common/Macrobench.tsx
frouioui: This is a good point, I am not an expert of TypeScript but I gave it a try on the three `Row` functions we have: [10a00d9](https://github.com/vitessio/arewefastyet/pull/517/commits/10a00d970c2a6eaec3bf158a869a84876c9e664c), [dc4199f](https://github.com/vitessio/arewefastyet/pull/517/commits/dc4199f1b09e58dfa2551a5083778398e5e69ef0) and [058af50](https://github.com/vitessio/arewefastyet/pull/517/commits/058af50dce9554f8b8cb4f7dea40426c41309893)

vmg (approved):
This looks great to me -- I've only reviewed the Golang code because the React stuff is above my paygrade. I don't think we should block on anything on merging this: the ultimate review will be testing it in practice with real benchmarks. There may be issues that simply cannot be seen by reading the code.

Overall, fantastic work here flo

Issue #500, opened 2024-01-23
Arewemobileyet: make this site mobile responsive


Maniktherana:
**Is your feature request related to a problem? Please describe.**
Currently, this site isn't mobile responsive. Accommodating mobile screens is a must on all websites. Critical information may be hidden when viewing from smaller viewports. 

![image](https://github.com/vitessio/arewefastyet/assets/14011425/de9142c8-0761-44e9-9a7c-de7d499059c4)![image](https://github.com/vitessio/arewefastyet/assets/14011425/beae0547-16d5-4391-9717-d539822b8103)
**Describe the solution you'd like**
This will involve adding media queries and possibly separate, mobile-only components such as a mobile navbar.

marsian83:
Hi @Maniktherana, you may check the responsiveness in #469 
The PR has not yet been merged as some minor changes are to be made still 
But you may check if there are any missing portions of responsive behavior in that PR's head

Maniktherana:
@marsian83 Ah didnt realize, I'll check it out

frouioui:
this has been fixed/implemented by @Jad31

PR #526, opened 2024-04-22
feat: make the table of status page responsive

utnim2:
Make the status table mobile responsive

[Screencast from 23-04-24 02:29:13 AM IST.webm](https://github.com/vitessio/arewefastyet/assets/127925465/0a66ad01-27cd-4f88-bcc4-dfe8d6fa83c0)

Related to #500

### DIFF
--- a/website/src/common/DisplayList.tsx
+++ b/website/src/common/DisplayList.tsx
@@ -44,7 +44,7 @@ export default function DisplayList(props) {
             <tr>
               {Object.keys(data[0]).map((item, key) => (
                 <th
-                  className="border-b min-w-[150px] text-left font-semibold py-2 border-front"
+                  className="border-b min-w-[184px] text-left font-semibold py-2 border-front"
                   key={key}
                 >
                   {item}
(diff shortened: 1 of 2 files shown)


frouioui (changes requested):
Thank you for your first contribution! The mobile version looks great to me! However, on a large enough screen the content is not centered properly and was moved to the left:
image

Moreover, we talked about this offline as you mentioned we should add a third (medium/tablet) size, because right now there is a weird interval when the window is between big and small the content is not responsive at all.

utnim2:
> Thank you for your first contribution! The mobile version looks great to me! However, on a large enough screen the content is not centered properly and was moved to the left: 
>
Fixed in 
> Moreover, we talked about this offline as you mentioned we should add a third (medium/tablet) size, because right now there is a weird interval when the window is between big and small the content is not responsive at all.

fixed in

frouioui (approved):

frouioui:
This looks good to me! Thank you for the contribution 🙇🏻

PR #528, opened 2024-04-26
feat: added shadcn `table` component

Gmin2:
Added the shadCn table component in DisplayList.tsx

### DIFF
--- a/website/src/common/DisplayList.tsx
+++ b/website/src/common/DisplayList.tsx
@@ -14,6 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import React from "react";
+import {
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from "@/components/ui/table";
 
 export default function DisplayList(props) {
   const { data } = props;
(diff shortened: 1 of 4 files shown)


website/src/components/ui/table.tsx
```diff
@@ -0,0 +1,117 @@
+import * as React from "react"
```
frouioui, line 1: We need a licence header there

website/src/common/DisplayList.tsx
```diff
@@ -39,33 +47,24 @@ export default function DisplayList(props) {
...
-        </table>
+            </TableRow>
+          </TableHeader>
+          <TableBody>
+            {data.map((val, key) => (
+              <TableRow key={key} className="hover:bg-accent cursor-default">
+                {Object.keys(val).map((item, key) => (
+                  <TableCell key={key}>{val[item]}</TableCell>
+                ))}
+              </TableRow>
```
frouioui, line 64: I think to go one step further we could also replace the "badge" we are using to show the benchmark's status: finished, failed, started. [We currently use pure CSS ](https://github.com/vitessio/arewefastyet/blob/c36d29a26e5b5a3bf62f97b9292891a4c1b718d3/website/src/pages/StatusPage/components/PreviousExecutions.tsx#L70-L79)but we could use Shadcn too https://ui.shadcn.com/docs/components/badge/

frouioui (commented):
So far it looks good to me, I left a couple comments.

What do you think about modifying the responsive (mobile) version of the table to use shadcn too?

utnim2:
> What do you think about modifying the responsive (mobile) version of the table to use shadcn too?

i dont think mobile view will be suited for this page(the shadcn component) as it shrinks and make the text size small it will be difficult to see the data .

image

frouioui:
> i dont think mobile view will be suited for this page(the shadcn component) as it shrinks and make the text size small it will be difficult to see the data

We could however have a card that contains the information we need

website/src/components/ui/badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+    },
+  }
+)
+
+export interface BadgeProps
+  extends React.HTMLAttributes<HTMLDivElement>,
+    VariantProps<typeof badgeVariants> {}
+
+function Badge({ className, variant, ...props }: BadgeProps) {
+  return (
```
frouioui, line 47: This is great thank you. Although I think we should just return nothing or an empty prop if the variant is empty. For instance in the execution queue we don't have any status and it shows like this right now:
image

frouioui: I think ultimately we should edit the table used by the execution queue to only show data we care about. And remove the columns: `Status`, `UUID`, `Started At`, `Finished At`, `Go Version`.

website/src/components/ui/badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+  {
+    variants: {
+      variant: {
+        default:
+          "border-transparent bg-primary text-primary-foreground hover:bg-primary/80",
+        secondary:
+          "border-transparent bg-secondary text-secondary-foreground hover:bg-secondary/80",
+        destructive:
+          "border-transparent bg-destructive text-destructive-foreground hover:bg-destructive/80",
+        outline: "text-foreground",
```
frouioui, line 33: I think the previous colors (Green, Red, Blue) were easier to distinguish the status of the benchmark. Let's switch back to those primary colors.

frouioui (commented):
This is really good job, thank you 🙏🏻 Just a few more comments.

utnim2:
> We could however have a card that contains the information we need

do we need this @frouioui ?

frouioui:
Hello! This change has recently been implemented by @Jad31, i will close this PR for now.

Issue #539, opened 2024-04-30
🐛 Bug: Linting and formating problems 


DhairyaMajmudar:
### Describe the bug

in the files of website folder, there are several linting  and formatting problems which are needed to be solved. 

### Steps To Reproduce

Open and code files changes can be seen there.

### Expected Behavior

There should be a proper linting and formating of all code files 

### Screenshots

n/a

### Device Information [optional]

```markdown
- OS:Debian (Linux)
- Browser: Firefox
- version: Latest
```

### Are you working on this issue?

Yes

DhairyaMajmudar:
I have tried solving this problem.
 
Pls. once have a look at the attached PR.

PR #541, opened 2024-05-01
Fix: Fixing formating and linting across the website code

DhairyaMajmudar:

Formats the code files.

**Issue Number:**
#539 

**Screenshots/videos:**
N/A

N/A

**Summary**
Code files are now formatted


utnim2:
In my opinion we should add  a script here that enforces linting and check on every PR through github actions, 
And mentioned it in the contribuding guide
In this way we dont have to do this manually all the time, the actions will fail if the changes pushed does not have linting

utnim2:
In my opinion we should add  a script here that enforces linting and check on every new PR created through github actions, 
And mentioned it in the contribuding guide
In this way we dont have to do this manually all the time, the actions will fail if the changes pushed does not have linting

frouioui:
I think we should do what @utnim2 mentioned

DhairyaMajmudar:
Sure will work on this 💯

DhairyaMajmudar:
@frouioui I have added the script `npm run format` using which the files can be formatted. Also updated the readme.md file : )

DhairyaMajmudar:
> To come back on what @utnim2 was saying, I think we should make this a CI check on every PR just like what we have in `.github/workflows/golang_lint.yaml`.

Sure looking forward to add lint CI

.github/workflows/typescript_lint.yaml
```diff
@@ -0,0 +1,53 @@
+name: PR Workflow
```
frouioui, line 1: The name of the workflow should be more descriptive 😄

.github/workflows/typescript_lint.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    paths:
+      - "website/**"
+
+jobs:
+  linting_and_type-checking:
+    name: Linting, Formatting and Type checking
```
frouioui, line 11: The job name as well, we should mention it is related to react/ts

.github/workflows/typescript_lint.yaml
DhairyaMajmudar: Yeah good catch : D 

I have improved this in my recent commit.

.github/workflows/typescript_lint.yaml
DhairyaMajmudar: Done this one also : )

DhairyaMajmudar:
Moving changes to PR https://github.com/vitessio/arewefastyet/pull/545

PR #547, opened 2024-05-22
Fix: NavBar for tablet and split screen view 

Ari1009:
### Description

Just a small fix for the navigation bar when in split-screen or tablet view. Previously, the text was overlapping.

### Screenshots

 Before and After

image


Ari1009:
@frouioui can you give a review?

PR #560, opened 2024-06-25
Benchmark the last three major releases of vitess

frouioui:
This PR enables benchmarks for the last three major releases of Vitess. It was previously set to `-1`, then changed to `-2` in https://github.com/vitessio/arewefastyet/pull/559.


frouioui:
This won't be done, `- 2` already englobes the last known major release of vitess and the two previous, thus covering all supported releases of vitess (the last three)

Issue #558, opened 2024-06-19
(Figma Design) UI Enhancements and New Features


Jad31:
## Description

This issue outlines a series of enhancements and new features to improve the user interface and functionality based on the [Figma](https://www.figma.com/design/kPIMCP3A5hQeD66UTKtuQH/arewefastyet?node-id=14-21&t=KUcTq5bj076374Uo-1)  design specifications. The changes include updates to existing pages, new pages, and various other improvements for a more streamlined and user-friendly experience.

### Home Page

- [ ] OK

### Status Page

- [ ] Reduce the table width
- [ ] Modify Hero description
- [ ] Reduce the size of metrics cards and their text

### Daily Page

- [ ] Align Hero section with Status Page
- [ ] Reuse the same chart card as on the Home Page
- [ ] Make the first chart card selected by default on page load

### Compare Page

- [ ] Remove Macro Benchmark title
- [ ] Add Hero section
- [ ] Add options to choose between releases, SHA, and commits for old and new
- [ ] Table:
  - [ ] Align Metrics column to the right
  - [ ] Center-align all other columns
  - [ ] Add tooltips on hover for badges in the delta column (Significant/No Significant)
  - [ ] Add vertical line between Metrics and old column
  - [ ] Add vertical line between new and P column
  - [ ] Group P and delta columns closely
  - [ ] Make QPS category including Read, Write, and others
  - [ ] Make Total CPU category including vtgate and vttablet
  - [ ] Make Total Allocated category including Read, Write, and others
  
### Foreign Keys
  
  - [ ] @frouioui is considering changes 🤔

### Pull Request

- [ ] Use the same date UI for "Opened at" as on the Compare Page table
- [ ] Refactor UI with Shadcn design

### Pull Request/{prNumber}

  - [ ] @frouioui is considering changes 🤔

### [Compare Query Plans](https://benchmark.vitess.io/macrobench/queries/compare?ltag=4f04b8b11e8b1dbc12aa187a5f53e8ed3b69b3f1&rtag=4f04b8b11e8b1dbc12aa187a5f53e8ed3b69b3f1&type=OLTP)

- [ ] Replace A and B labels with Old and New respectively
- [ ] Update Hero section to match other pages
- [ ]  Refactor modal using Shadcn design

### History

- [ ] Add Hero section
- [ ] Table:
  - [ ] First column: SHA
  - [ ] Second column: Source
  - [ ] Third column: Workload (number of benchmarks)
[...]

Issue #525, opened 2024-04-22
LFX third iteration of UI improvement: Shadcn, Typescript and Responsive


frouioui:
## What is arewefastyet

Arewefastyet is the automated and continuous benchmarking platform for Vitess (https://github.com/vitessio/vitess). It automatically performs different types of benchmarks on each new commit on `main` and release branches, on new tags and releases, and on selected Pull Request in https://github.com/vitessio/vitess. It can run hundreds of benchmark a day, allowing maintainers and adopters to know if a certain version of Vitess is slower or faster than another.

To dive deeper into the internals of arewefastyet you can refer to the following talks. Although they are a little bit old and the internal of how we benchmark changed, these talks will give you a good overview of what arewefastyet is doing.
 
- https://videos.perconalive.com/watch/bPeJAqQ5FomWsJvnehKYoH
- https://www.youtube.com/watch?v=h1bd1yxA4ec
- https://www.youtube.com/watch?v=fLullHPddvg

## Previous LFX terms

- Iteration 1 (Jun-Aug 2023): https://github.com/vitessio/arewefastyet/issues/328
- Iteration 2 (Sept-Nov 2023): https://github.com/vitessio/arewefastyet/issues/415 

### Expected Work

The mentee and I will begin this term by working on Figma for a little bit to think about the future design of the website, there is about 10 pages to design. We will build a Figma using Shadcn components. The mentee will then use this Figma board to revamp the website page by page. The new design is expected to be responsive. While revamping the website, it is expected to finish migrating to TypeScript and type as much components as possible (as we see fit).

Purely informative examples of things we want to change:
- [The status page](https://benchmark.vitess.io/status) is not responsive at all today. We can also improve the top-part of the page with more information using a [dashboard like view](https://ui.shadcn.com/examples/dashboard). The table in the bottom-part could be modified to use a more sober design as in https://ui.shadcn.com/docs/components/table. In the same table, for benchmark in progress, we could have a small animation.
- On all pages the loader animation could be changed to something else.
- [The comparison page](https://benchmark.vitess.io/macro?old=19.0.0&new=19.0.3) has some rendering issues for some of the tables, and is also not responsive and contains a lot of information per table.

If the revamp of the website is completed before the end of the internship, the mentee and mentor will work on adding more features to the website: filtering of the tables in the [status page](https://benchmark.vitess.io/status), add password-protected pages for our admin API endpoints, add Slack notifications about benchmark runs.

### Skills required

[...]

PR #562, opened 2024-07-02
Refactor navbar

Jad31:

Refactor the navbar to match the new figma design.
Fix responsive issues as well

**Issue Number:**

-  Closes #___ 
-  Related to #558  
-  Others? 

**Screenshots/videos:**

|  Before | After |
|  -----  | ----- |
|         |       |   

**Summary**

### DIFF
--- a/website/src/common/Navbar.tsx
+++ b/website/src/common/Navbar.tsx
@@ -48,6 +48,7 @@ const navItems: NavItem[] = [
   { to: "/macro", title: "Macro" },
   { to: "/fk", title: "Foreign Keys" },
   { to: "/pr", title: "PR" },
+  { to: "/history", title: "History" },
 ];
 
 export default function Navbar(): JSX.Element {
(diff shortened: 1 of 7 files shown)


website/src/pages/HistoryPage/HistoryPage.tsx
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright 2023 The Vitess Authors.
```
frouioui, line 2: nit: when creating new files, we should update the license header to the current year.

```suggestion
Copyright 2024 The Vitess Authors.
```

website/src/pages/HistoryPage/components/Hero.tsx
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright 2023 The Vitess Authors.
```
frouioui, line 2: Same here

```suggestion
Copyright 2024 The Vitess Authors.
```

website/src/pages/HistoryPage/components/Hero.tsx
frouioui, line 1: This is already planned for a follow-up PR already, but putting it out there so it is known: let's extra the Hero into its own component that we will re-use on every page, that will ensure the Hero is consistent everywhere and will make the codebase more maintainable.

website/src/common/Navbar.tsx
frouioui, line 1: It would look nice to add a bit more space between each links on the nav bar, for now it looks really close, making it harder to read. I think using `md:gap-x-3` instead of `md:gap-x-2` looks slightly better without compromising the responsiveness.

**Before (md:gap-x-2)**
image

**After (md:gap-x-3)**
image

frouioui (approved):
Left some comments, otherwise this is a nice addition, thank you!

FYI, this was the design:
image

And this is the current result:
image

The search bar will be added in a follow-up PR when we work on the History page.

The responsiveness works well too.

frouioui:
Linked to https://github.com/vitessio/arewefastyet/issues/525

frouioui (approved):
LGTM, nice first contribution! 🙇🏻

PR #568, opened 2024-07-09
Refactor Daily Page

Jad31:

**Issue Number:**

-  Closes #___ 
-  Related to #___ 
-  Others? 

**Screenshots/videos:**

|  Before | After |
|  -----  | ----- |
|         |       |   

**Summary**


website/src/pages/DailyPage/components/DailyCharts.tsx
```diff
@@ -0,0 +1,307 @@
...
+  const { benchmarkType } = props;
+  const workloads: Workloads[] = [
+    "OLTP",
+    "OLTP-READONLY",
+    "OLTP-SET",
+    "TPCC",
+    "TPCC_FK",
+    "TPCC_UNSHARDED",
+    "TPCC_FK_UNMANAGED",
+  ];
```
frouioui, line 104: I don't think we should hardcode this.

website/src/pages/DailyPage/components/DailyDailySummary.tsx
```diff
@@ -0,0 +1,87 @@
+/*
+Copyright 2023 The Vitess Authors.
```
frouioui, line 2:
```suggestion
Copyright 2024 The Vitess Authors.
```

website/src/pages/DailyPage/components/DailyCharts.tsx
```diff
@@ -0,0 +1,307 @@
...
+  data: ChartDataItem[];
+  title: string;
+  colors: string[];
+};
+
+const chartConfig: { [key: string]: { label: string; color: string } } = {
+  qpsReads: {
+    label: "Reads",
+    color: "hsl(var(--chart-qpsReads))",
+  },
```
frouioui, line 47: I am guessing the keys here are what we should change in order to have proper names in the legend?

frouioui: Ah nevermind, I think it is in `chartMetadas`

website/src/pages/DailyPage/components/DailyCharts.tsx
```diff
@@ -0,0 +1,307 @@
...
+  const {
+    data: dataDaily,
+    error: dailyError,
+    textLoading: dailyTextLoading,
+  } = useApiCall<MacroData>(
+    `${import.meta.env.VITE_API_URL}daily?workloads=${workloadsQuery}`
+  );
+
+  const chartData = dataDaily.map((item) => ({
+    gitRef: item.git_ref.slice(0, 8),
```
frouioui, line 117: Let's have a common function that returns the short version of a Git SHA. That way we can call this function everywhere and if someday we want to change from 8 chars to 7, or 6, it will be easy to do.

website/src/pages/DailyPage/components/DailyCharts.tsx
```diff
@@ -0,0 +1,307 @@
...
+    "OLTP",
+    "OLTP-READONLY",
+    "OLTP-SET",
+    "TPCC",
+    "TPCC_FK",
+    "TPCC_UNSHARDED",
+    "TPCC_FK_UNMANAGED",
+  ];
+
+  const workloadsQuery = workloads.join("&workloads=");
```
frouioui, line 106: To match the proposed change to `api.go`:

```suggestion
  const workloadsQuery = workloads.join("&workload=");
```

website/src/pages/DailyPage/components/DailyCharts.tsx
frouioui: Actually, looking at this again, we should not send all the workloads to the API. Only one is enough, since we are only displaying a single workload. Currently, we send all the workload no matter what workload was selected by the user. And we end up only looking at the data from the first element of the response (`OLTP`), even if the user selected i.e. `TPCC`. You can reproduce this by going to http://localhost/daily?type=TPCC_FK_UNMANAGED, the numbers/results will be the same as the `OLTP` workload.

frouioui: Actually I am not sure what workload's results we end up using, but this is wrong.

website/src/pages/DailyPage/components/DailyDailySummary.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const { benchmarkType, setBenchmarktype } = props;
+  const workloads: Workloads[] = [
+    "OLTP",
+    "OLTP-READONLY",
+    "OLTP-SET",
+    "TPCC",
+    "TPCC_FK",
+    "TPCC_UNSHARDED",
+    "TPCC_FK_UNMANAGED",
+  ];
```
frouioui, line 37: We should not hardcode this part.

go/server/api.go
```diff
@@ -51,6 +51,11 @@ type RecentExecutions struct {
 	FinishedAt    *time.Time `json:"finished_at"`
 }
 
+func (s *Server) getWorkloadList(c *gin.Context) {
+	slog.Info("getWorkloadList", s.benchmarkTypes);
```
frouioui, line 55: We can remove this I think

website/src/common/DailySummary.tsx
```diff
@@ -60,8 +51,18 @@ export default function DailySummary({ data }: DailySummaryProps) {
     }));
   }
 
+  const getBenchmarkType = () => {
```
frouioui, line 54: let's use workload here too

website/src/pages/DailyPage/components/DailyCharts.tsx
```diff
@@ -0,0 +1,316 @@
+import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
+import { ChartContainer, ChartTooltip } from "@/components/ui/chart";
```
frouioui, line 2: License header

PR #587, opened 2024-07-25
Feat search

Jad31:

**Issue Number:**

-  Closes #___ 
-  Related to #___ 
-  Others? 

**Screenshots/videos:**

|  Before | After |
|  -----  | ----- |
|         |       |   

**Summary**

### DIFF
--- a/website/src/common/CommandMenu.tsx
+++ b/website/src/common/CommandMenu.tsx
@@ -14,15 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { type DialogProps } from "@radix-ui/react-dialog";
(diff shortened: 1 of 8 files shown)


website/src/common/CommandMenu.tsx
frouioui, line 1: Just one thing that does not work is:
- I am on the "History" page
- I press CMD+K and search "toto"
- That works, we can "toto" in the search bar of the table.
- When I press CMD+K again and search "tata"
- I can see the URL changes to use the proper `gitRef` value, but the search bar of the table does not change

frouioui (commented):
Besides that bug I wrote, it looks good to me! thank you 🚀

frouioui (approved):
This is great! Thank you 🚀

PR #604, opened 2024-08-27
fix: check both maintainer and arewefastyet team admin

Jad31:

**Issue Number:**

-  Closes #___ 
-  Related to #___ 
-  Others? 

**Screenshots/videos:**

|  Before | After |
|  -----  | ----- |
|         |       |   

**Summary**

### DIFF
--- a/go/admin/api.go
+++ b/go/admin/api.go
@@ -126,21 +126,25 @@ func (a *Admin) handleGitHubCallback(c *gin.Context) {
 }
 
 func (a *Admin) checkUserOrgMembership(client *goGithub.Client, username, orgName string) (bool, error) {
-	teams, _, err := client.Teams.ListTeams(context.Background(), orgName, nil)
-	if err != nil {
-		return false, err
-	}
-	for _, team := range teams {
-		if team.GetName() == maintainerTeamGitHub || team.GetName() == arewefastyetTeamGitHub {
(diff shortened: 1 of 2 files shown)


go/admin/api.go
```diff
@@ -126,21 +126,25 @@ func (a *Admin) handleGitHubCallback(c *gin.Context) {
...
-				}
-				return false, err
-			}
-			return membership.GetState() == "active", nil
-		}
-	}
-	return false, nil
-}
+    teams, _, err := client.Teams.ListTeams(context.Background(), orgName, nil)
+    if err != nil {
```
frouioui, line 130: Let's keep using tabs where we can

main
frouioui, line 1: Should be removed

go/admin/api.go
```diff
@@ -126,21 +126,25 @@ func (a *Admin) handleGitHubCallback(c *gin.Context) {
...
+            membership, _, err := client.Teams.GetTeamMembership(context.Background(), team.GetID(), username)
+            if err != nil {
+                if strings.Contains(err.Error(), "404 Not Found") {
+                    continue
+                }
+                return false, err
+            }
+            if membership.GetState() == "active" {
+                isMember = true
+            }
```
frouioui, line 146: we could even break here and avoid having to do a second API call

frouioui (approved):
Looks good to me, approving and merging later in case you want to adresse the feedback

PR #627, opened 2024-11-19
Allow Admin to add PR to execution queue for benchmarking and profiling

harshit-gangal:
This PR adds to Admin page a feature to allow Admins to send PR number for benchmarking instead of SHA

### DIFF
--- a/go/admin/templates/add_new_executions.html
+++ b/go/admin/templates/add_new_executions.html
@@ -20,6 +20,15 @@
         name="sha"
       />
     </div>
+    <div>
+      <label class="label">PR</label>
+      <input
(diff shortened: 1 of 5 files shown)


frouioui (approved):

go/admin/templates/add_new_executions.html
```diff
@@ -20,6 +20,15 @@
         name="sha"
       />
     </div>
+    <div>
+      <label class="label">PR</label>
```
frouioui, line 24:
```suggestion
      <label class="label">PR Number</label>
```

frouioui (approved):

PR #678, opened 2026-05-12
Pin GitHub Actions to commit SHAs

mhamza15:
## What's this?
Pins the GitHub Actions used by these workflows to the commit SHAs currently targeted by their existing refs.

## How it works
This keeps the same action versions and replaces tag or branch refs with exact SHAs. No action upgrades here.

Note: `ansible/ansible-lint-action@main` is intentionally left alone for now.


timvaillancourt (approved):
🔒 🚀

arthurschreiber (approved):

arthurschreiber:
> Note: `ansible/ansible-lint-action@main` is intentionally left alone for now.

This is causing builds to fail now. Should we change this too? I'm curious why this wasn't changed as part of this PR. 🤔

mhamza15:
> > Note: `ansible/ansible-lint-action@main` is intentionally left alone for now.
> 
> This is causing builds to fail now. Should we change this too? I'm curious why this wasn't changed as part of this PR. 🤔

Yep we should. I/Codex missed it initially because the repo doesn't exist anymore or was renamed it seems. I believe this is the one now: https://github.com/ansible/ansible-lint. Will update

mhamza15:
Okay changed to https://github.com/ansible/ansible-lint v25. v26 is actually the latest, but it uses `astral-sh/setup-uv` internally without a pinned SHA, so it trips our policy as well. Ideally we should just run the tool directly with our own composite action.

The job still fails because of tons of linting errors, but leaving that out of scope of this PR for now.

PR #694, opened 2026-06-18
Fix failing CI: Golang Lint and Ansible Lint

arthurschreiber:
## Summary

Two CI workflows have been red on `main` for a while due to **tooling/version bumps flagging pre-existing code** — unrelated to any recent feature work. This PR makes both green.

### 1. Golang Lint (`golangci-lint` v1.60.1)
Bumping golangci-lint enabled newer `errcheck`/`staticcheck` checks (and a `govet` analyzer that chokes on generics):
- **errcheck** — handle previously-unchecked error returns (deferred `Close()` on `sql.Rows`/response bodies, `os.RemoveAll` in tests, `fmt.Fprint*`).
- **staticcheck** — QF1008 (redundant embedded `Mode` selector), QF1004 (`strings.ReplaceAll`), QF1001 (De Morgan), QF1002 (tagged switch), ST1005 (error-string capitalization/newlines).
- **govet** — a minimal `.golangci.yml` filters the false positive `cannot inline: type parameter inference is not yet supported` on generic stdlib helpers (`slices.Contains`). Default linter set otherwise unchanged.

### 2. Ansible Lint (`ansible-lint` v25.12.2)
The action was pinned to v25.12.2, whose ruleset reports ~660 violations (the old numeric `-x` skips no longer apply):
- Auto-fixed the **mechanical** violations via `ansible-lint --fix` — FQCN module names, YAML formatting, task/play naming, Jinja spacing.
- Converted role-include `vars:` blocks from **list form to mappings** (fixes the unskippable `syntax-check`/`schema` errors).
- Added missing task/play names.
- Added `.ansible-lint` skipping rules that would require **cross-role variable renames or behavioral changes** to the production benchmarking playbooks (`var-naming`, `no-changed-when`, `ignore-errors`, `command-instead-of-*`, `no-free-form`, `risky-*`), plus `yaml[line-length]` (previously skipped as the old `-x 204`). These are intentionally deferred because they can't be validated without running the real automation.

## Verification
Both linters were run locally at the exact CI versions:
- `golangci-lint`: all `errcheck`/`staticcheck` cleared; only the excluded `govet` generics false-positive remains; `go build ./go/...` passes.
- `ansible-lint -x 301,302,303,305,306,204,208 ansible/`: **0 failures, 0 warnings** (reaches the `production` profile).

## Not included
- `var-naming` and behavioral ansible rules (skipped, see above) — would be a separate, infra-tested change.
- The repo's other red checks on feature PRs (the `@v3`/`@v4` SHA-pin policy errors, stale CLI docs) are branch-rebase artifacts, not `main` breakages, and aren't touched here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


arthurschreiber:
Split into two separate PRs: #695 (Golang Lint) and #696 (Ansible Lint).

